PR #1, opened 2022-06-07
DSEGOG-3 DSEGOG-5 DSEGOG-6 Adding initial basic table and column functionality 

sam-glendenning:
## Description
This includes the use of React Table for the headless table functionality and a small amount of Material UI to make it look somewhat presentable. Right now this focuses on three main pieces of functionality:

1. Display of data in a table (data itself is randomly generated locally for now while we don't have an API)
2. Table pagination
3. Table column sorting (currently columns are multi-sorted but this will likely be changed to be single sort as multi-sort is not a specific OG requirement)

## Testing instructions
Double check that you are set up with yarn 2 and that it can zero-install all the required packages with `yarn install`. This should be an automatic process thanks to the `.yarnrc.yml` file. You should also be able to open the project correctly in VS Code (there is a configuration for this in `.yarn/sdks`.
Then try out the table in the browser (data display and pagination should work fine and sorting a column will result in the column header showing as sorted but not actually manipulating the table). Also make sure adding and removing columns by ticking and unticking the checkboxes works as expected. The checkboxes should refresh with what data is available based on what page you're on.

## Agile board tracking
Closes [DSEGOG-3](https://stfc.atlassian.net/browse/DSEGOG-3), [DSEGOG-5](https://stfc.atlassian.net/browse/DSEGOG-5), [DSEGOG-6](https://stfc.atlassian.net/browse/DSEGOG-6)


src/index.tsx
```diff
@@ -28,17 +28,22 @@ function domElementGetter(): HTMLElement {
...
+// Defining loadRootComponent did not throw this warning
+// May be worth investigating further
 const reactLifecycles = singleSpaReact({
   React,
   ReactDOM,
-  rootComponent: App,
+  // rootComponent: App,
   domElementGetter,
+  loadRootComponent: () =>
+    new Promise((resolve, reject) => resolve(() => <App />)),
```
louise-davies: I don't see why we need to use `loadRootComponent` here - and I think it would be good as well to move the error boundary stuff from `App` to here

louise-davies (changes requested):

src/index.tsx
sam-glendenning: @louise-davies Made the changes

louise-davies (approved):
Looks good - I've not gone over all the WIP code with a fine tooth comb as I know most of it is subject to change but it looks fine and works well so far :)

PR #2, opened 2022-06-09
DSEGOG-9 user can remove columns with close icon and middle mouse click

sam-glendenning:
## Description
The user is now able to more easily remove columns from display. A black cross icon is displayed in each column header which the user can click to remove the column. I also added in the ability for the user to click the entire column header with the middle mouse button as an alternative method for removing the column from display. This mimics the behaviour of popular web browsers, where a user can middle click a tab to close it. This may be a familiar action for some users and speed up their interaction with the site.

## Testing instructions
Try removing columns using both methods and test that the functionality works as expected. Try this across multiple pages to double check it's working. Also make sure that, when a column is closed, its checkbox is unticked at the bottom of the page.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes [DSEGOG-9](https://stfc.atlassian.net/jira/software/projects/DSEGOG/boards/22?selectedIssue=DSEGOG-9)

### DIFF
--- a/src/table/headerRenderers/dataHeader.component.test.tsx
+++ b/src/table/headerRenderers/dataHeader.component.test.tsx
@@ -13,7 +13,15 @@ describe('Data Header', () => {
   const onClose = jest.fn();
 
   const createView = (): RenderResult => {
-    return render(<DataHeader {...props} />);
+    return render(
+      <table>
+        <thead>
+          <tr>
+            <DataHeader {...props} />
+          </tr>
+        </thead>
+      </table>
+    );
   };
 
   beforeEach(() => {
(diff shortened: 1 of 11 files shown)


louise-davies (changes requested):
Code looks good - I don't like that there's an extra floating line under the headers (see screenshot) - can you try and remove that? I think the issue is that there's two `TableCell`s

image

src/table/headerRenderers/dataHeader.component.test.tsx
```diff
@@ -13,7 +13,15 @@ describe('Data Header', () => {
...
-    return render(<DataHeader {...props} />);
+    return render(
+      <table>
+        <thead>
+          <tr>
+            <DataHeader {...props} />
+          </tr>
+        </thead>
+      </table>
+    );
```
sam-glendenning, line 24: This tests the column header in a proper tabular context while also removing the error of invalid DOM nesting in snapshot tests

louise-davies (approved):

PR #4, opened 2022-06-14
DSEGOG-7 user can resize columns

sam-glendenning:
## Description
Adding functionality to allow the user to resize the table columns by dragging the dividers placed between column headers. This uses React Table's useResizeColumns plugin to natively handle table resizing and management of column state.

The functionality works reasonably well but may need to be expanded upon later. One notable bug that exists at the moment is that, when the columns are reordered, added to or removed from, the sizes go back to the default width. We'll need to fix this later on. However, for the meantime, this might satisfy the requirement.

This PR also includes the config file defining the types necessary to extend the functionality of React Table to include all of its plugins. Tanstack do not maintain these types, hence we define them ourselves. See here for more info - https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table

## Testing instructions
Experiment with resizing columns as well as adding and removing columns to see how the functionality works together.

- [ ] Review code
- [ ] Check Actions build
- [ ] Review changes to test coverage

## Agile board tracking
Closes [DSEGOG-7](https://stfc.atlassian.net/jira/software/projects/DSEGOG/boards/22?selectedIssue=DSEGOG-7)

PR #5, opened 2022-06-16
DSEGOG-8 user can reorder columns

sam-glendenning:
## Description
Note: this contains resizing columns functionality so probably best to review that PR first - https://github.com/ral-facilities/operationsgateway/pull/4

Adding functionality to allow the user to reorder the table columns using drag and drop functionality provided by [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd). The result is a nice looking, simplistic method by which the user can reorder the displayed columns without having to add or remove columns to achieve the same thing. To drag and drop, the user can click anywhere on the column header that isn't the close button or the divider. I had to be a bit clever to make sure their functionality didn't overlap.

It's not really feasible to test this functionality outside of a browser so I used this opportunity to add Cypress tests to the project. I've only done tests for this functionality so far, not made any for the previous functionality I've done yet. Currently using Cypress 9.6.0 to match DataGateway because this was the simplest to get set up with. I considered upgrading to Cypress v10 but this contains many breaking changes so I made a separate issue (stfc.atlassian.net/jira/software/projects/DSEGOG/issues/DSEGOG-33) for this to tackle later.

## Testing instructions
Experiment with adding columns to the display and then reordering them by dragging the headers. Experiment with both large and small tables to make sure the functionality works in all cases. Also make sure that the drag and drop functionality does not interfere with column resizing and closing.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes [DSEGOG-8](https://stfc.atlassian.net/jira/software/projects/DSEGOG/boards/22?selectedIssue=DSEGOG-8)


src/table/headerRenderers/dataHeader.component.tsx
```diff
@@ -84,58 +92,82 @@ const DataHeader = (props: DataHeaderProps): React.ReactElement => {
...
+              style={{
+                display: 'flex',
+                flexDirection: 'row',
+              }}
+              onMouseOver={() => {
+                setPermitResizing(false);
+              }}
+              onMouseOut={() => {
+                setPermitResizing(true);
+              }}
```
sam-glendenning: Most of the loss of codecov here is over these two events. Any advice on how to test these (or even if we should) is appreciated.

src/table/table.component.tsx
```diff
@@ -65,10 +76,54 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
+      return column.Header?.toString() ?? '';
+    });
+  };
 
-  const { getTableProps, getTableBodyProps, headerGroups, rows, prepareRow } =
-    tableInstance;
+  const handleOnDragEnd = (result: DropResult): void => {
+    if (!result.destination) return;
+    setColumnOrder(updater(result));
+  };
```
sam-glendenning: Again, loss of codecov here. Not even sure how to trigger this behaviour in a unit test. Worst comes to worst we can override the codecov threshold and just merge in?

sam-glendenning:
Noticed a small bug with column ordering when columns are added/removed

sam-glendenning:
@louise-davies apologies for the length this PR has turned out as but it is now ready for review again. It includes a lot of optimisations that improve the maintainability of things going forward. 

The gist of what I've done since I marked this as draft is fixing a bug which meant, when removing a column and adding it back, it went back into the same place it was removed from instead of being added to the end of the table as expected. I also added some code that more efficiently manages the column order and the table state overall. Please try and test as much of the functionality as possible to make sure everything plays nicely with each other.

Although this is a lengthy PR, remember that much of it is also taken up by yarn package additions and a couple of big snapshots!

src/table/headerRenderers/dataHeader.component.tsx
```diff
@@ -87,64 +92,85 @@ const DataHeader = (props: DataHeaderProps): React.ReactElement => {
...
+            >
+              <div aria-label={`close ${dataKey}`}>
+                <StyledClose onClick={() => onClose(dataKey)} />
+              </div>
+              <div
+                {...resizerProps}
+                style={{
+                  marginLeft: 10,
+                  paddingLeft: '4px',
```
louise-davies: This I think was a merge conflict resolution gone wrong - as adding this back in has reverted the fix I made in the resizing PR. I shall fix this as I've already done the code - just need to update the snapshots in a sec

sam-glendenning: Thanks Louise, looks like I got confused

src/table/table.component.tsx
```diff
@@ -28,24 +36,37 @@ export interface TableProps {
...
     onSort,
-    onClose,
   } = props;
 
+  /*
+   ** A note about the columns used in this file:
+   ** availableColumns - these are the columns passed in from RecordTable.
+   **                    They represent all columns that can currently be added to the
+   **                    display based on data received from the backend
+   ** checkedColumns   - these are the columns the user has selected to appear in the table
```
louise-davies: Minor point, but I dislike the `checkedColumns` name as I think it's too based on the WIP column checkboxes selector, rather than what it will look like in the future. Maybe `selectedColumns` makes more sense? I'm not sure I care enough for you to change it though 😅

sam-glendenning: I also prefer selectedColumns. I can change in a later PR since I'm working on a new feature right now

src/table/table.component.tsx
```diff
@@ -78,13 +128,43 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
+    rows,
+    state,
+    prepareRow,
+    setColumnOrder,
+    visibleColumns,
+  } = tableInstance;
+
+  const { columnOrder } = state;
+
+  const updater = (result: any): string[] => {
```
louise-davies: The name of this function is very generic to the point it's hard to tell what it's for. Can we rename this something more verbose? colOrderUpdater or something??

louise-davies:
@sam-glendenning with the commit I've just done I just removed the right padding on table cells which fixes that overflow. I don't think we need the right padding but we can see later with real data if the text going right up to the border isn't wanted

src/table/table.component.tsx
```diff
@@ -168,6 +262,12 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
             page={page}
             rowsPerPage={resultsPerPage}
           />
+          <ColumnCheckboxes
+            availableColumns={availableColumns}
+            checkedColumns={checkedColumns}
+            onColumnOpen={handleColumnOpen}
+            onColumnClose={handleColumnClose}
+          />
```
louise-davies: Should this really be in the table component? In the future, this will be a separate epic and thus I don't think it should be in `Table`. However, by then I'd expect we'd have redux and perhaps that will result in this not having to be nested in `Table` as well...

sam-glendenning: My reasoning is that we can't have one without the other so they're intrinsically linked. If we have checkboxes, even if none of them are selected, we still have a table element and the table will always have an ID column loaded, for example

louise-davies: true, I suppose we'll see how the structure pans out when we redux everything & we refactor `ColumnCheckboxes` into the channel selector

louise-davies (approved):
Functionality works well and looks nice plus I can still access column closing & resizing.

PR #6, opened 2022-06-20
DSEGOG-10 sticky column headers and ID column, plus date/time selector

sam-glendenning:
## Description
This PR does a lot of things :

- made column headers sticky, such that when the table is scrolled vertically, they remain in the viewport. I've simulated this by increasing the number of results per page and setting a maximum height on the table container
- made the ID column sticky, such that when the table is scrolled horizontally, the entire column and header remains in the viewport. You can simulate this by adding more columns than can fit in the viewport together
- added a date/time filter option above the table, separate from the component itself. These contain two date ranges for both start date/time and end date/time. Whenever one is either successfully filled with a valid date or cleared, the onChange method is fired, which updates the record query
- made the ID column visible at all times, with no option to remove it as per the issue details
- made the ID column non-draggable since it always needs to be on the left of the table
- recreated a lot of tests based on the above ID column changes
- added some more Cypress tests and generally tried to bump the global coverage in unit tests. _Check snapshots carefully_ as there have been many changes made and we want to make sure we get the expected results

I also noticed a bug on Friday afternoon that was present before starting this issue. When a column is added or removed, the current widths of all columns are put back to their default, erasing all user customised table sizing. I have not issued a fix for this in this PR as it's late on Friday and I wanted to get this reviewed by the time I get from leave. Either we fix it here, if it's easily doable, or we create an issue for it and attack it later.

## Testing instructions
I've tried my best to make sure everything plays together via unit tests but the amount of features in this table is quickly stacking up and will continue to lead to more unintended consequences (see above note about column resizing bug). So please test new functionality with as much existing functionality as possible.

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes DSEGOG-10


src/table/table.component.tsx
```diff
@@ -151,29 +180,30 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
+              role="table-container"
+              component={Paper}
+              sx={{ maxHeight: 300, overflow: 'auto' }}
+            >
               <MuiTable {...getTableProps()}>
-                <MuiTableHead>
+                <MuiTableHead
+                  sx={{
+                    position: 'sticky',
+                    background: 'white', // theme colour later on
```
louise-davies: Can we add a TODO to this, as otherwise this comment could be forgotten about, whereas we know to search for TODO

louise-davies (commented):
Out of curiosity, can I ask why the date selector stuff was added to this PR? I think that stuff falls under the "Search" epic but if it's needed now then it's fine. Note that I think we only need two datepickers - as each shot has only 1 timestamp associated with it we just need a start and end picker for the range. This differs to DG where there were seperate columns in the DB which could have individual range filters applied to them

src/table/headerRenderers/dataHeader.component.tsx
```diff
@@ -91,84 +91,99 @@ const DataHeader = (props: DataHeaderProps): React.ReactElement => {
...
+          }}
+          onMouseOver={() => {
+            setPermitDragging(false);
+          }}
+          onMouseOut={() => {
+            setPermitDragging(true);
+          }}
+        >
+          {/* If this is an ID header, remove icon visibility but still render it
+          This ensures header widths remain consistent */}
```
louise-davies, line 166: The problem with this is that the ID column seemingly for no reason covers up the column name when resized very small. I've got a fix for this by using `display: none` and adjusting the `40px` value above

src/table/table.component.tsx
```diff
@@ -186,27 +216,42 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
                                 ref={provided.innerRef}
+                                sx={{
+                                  height: headerHeight, // Consistent height to account for headers that don't include any icons
+                                }}
                               >
                                 {headerGroup.headers.map((column, index) => {
                                   const { key, ...otherHeaderProps } =
                                     column.getHeaderProps();
 
+                                  const dataKey = column.render('id') as string;
```
louise-davies: I changed this in my other PR but I can ask it here - why do we use `column.render('id')` and not just `column.id`?

sam-glendenning: I _think_ that was a remnant from a tutorial I followed whereby they were rendering 'id' as a basic text element. But it makes sense here to just use column.id so I'll change it

sam-glendenning:
> Out of curiosity, can I ask why the date selector stuff was added to this PR? I think that stuff falls under the "Search" epic but if it's needed now then it's fine. Note that I think we only need two datepickers - as each shot has only 1 timestamp associated with it we just need a start and end picker for the range. This differs to DG where there were seperate columns in the DB which could have individual range filters applied to them

So I did it like this because I thought allowing for a range selector for the date was more convenient for the end user and the API would allow for it in the same way with less than and greater than comparisons. I didn't have a look at the tasks in the Search epic so was just copying DG's method. If this isn't reflective of what the final product will be, I can go back to a basic start and end date selection.

And as for why I did it in the first place - it was included in the issue title and description but there's more to it overall that means it probably belongs in another epic/issue. Will adjust what I have so far to include only basic date-time selection and merge that.

src/table/headerRenderers/dataHeader.component.tsx
sam-glendenning: That seems to work better, thanks for that

louise-davies:
@sam-glendenning We spoke about sticky columns in the quarterly report meeting and I got some opinions re: this requirement - see my comment here: https://stfc.atlassian.net/browse/DSEGOG-10?focusedCommentId=24900

Also, the merge conflicts need resolving 😅

sam-glendenning:
I will likely abandon this PR in favour of [this new one](https://github.com/ral-facilities/operationsgateway/pull/13) instead. I managed to completely bork the yarn.lock file when trying to fix merge conflicts caused by adding Renovate to the repo. Since the issue has been updated with [new information](https://stfc.atlassian.net/browse/DSEGOG-10?focusedCommentId=24900) anyway, I thought I might as well make a fresh PR to address those changes anyway.

I'll leave this open for now, with a view to close it if/when the other is merged.

louise-davies:
Closed in favour of #13

PR #7, opened 2022-07-01
DSEGOG-13 header tooltips

louise-davies:
## Description
Add tooltips to headers displaying some info on the channel upon hover. For now, I've added a TODO that we need to get the channel info from somewhere obviously. I know there's a point in the issue about conditionally showing system name in the tooltip but I feel that's easier to tackle once we're getting actual data channel data rather than our mocked column data. (maybe I should add a TODO to the code about that as well)

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Hover over some headers

## Agile board tracking
Closes DSEGOG-13

### DIFF
--- a/src/table/headerRenderers/dataHeader.component.test.tsx
+++ b/src/table/headerRenderers/dataHeader.component.test.tsx
@@ -42,10 +42,14 @@ describe('Data Header', () => {
       onClose: onClose,
       label: 'Test',
       icon: function Icon() {
-        return <div>Test</div>;
+        return <div>Icon</div>;
       },
       resizerProps: {},
       index: 0,
+      channelInfo: {
+        units: 'm',
+        description: 'test description',
+      },
     };
   });
 
@@ -137,5 +141,23 @@ describe('Data Header', () => {
       createView();
       expect(onSort).toHaveBeenCalledWith('test', 'desc');
     });
+  });
+
+  it('displays tooltip when user hovers over column name', async () => {
+    createView();
+    const header = screen.getByText('Test');
+
+    fireEvent(
+      header,
+      new MouseEvent('mouseover', {
+        bubbles: true,
+      })
+    );
+
+    expect(
+      await screen.findByText('Units: m', {
+        exact: false,
+      })
+    ).toBeInTheDocument();
   });
 });
--- a/src/table/headerRenderers/dataHeader.component.tsx
+++ b/src/table/headerRenderers/dataHeader.component.tsx
@@ -6,6 +6,7 @@ import {
   TableCell,
   SxProps,
   styled,
+  Tooltip,
 } from '@mui/material';
 import Close from '@mui/icons-material/Close';
 import React from 'react';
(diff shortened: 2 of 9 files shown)


src/table/headerRenderers/dataHeader.component.test.tsx
```diff
@@ -138,4 +142,22 @@ describe('Data Header', () => {
       expect(onSort).toHaveBeenCalledWith('test', 'desc');
     });
   });
+
+  it('displays tooltip when user hovers over column name', async () => {
+    createView();
+    const header = screen.getByText('Test');
```
sam-glendenning, line 148: I'm personally a fan of being more specific about the type of element we're trying to access. The `Tooltip` is wrapped around the header sort label, in which case I would put `screen.getByTestId('sort test')`. This allows you to be more descriptive about what part you're trying to test, compared to just referring to a piece of text that could appear anywhere on the screen

src/table/headerRenderers/dataHeader.component.tsx
```diff
@@ -110,27 +113,39 @@ const DataHeader = (props: DataHeaderProps): React.ReactElement => {
...
-                onMouseDown={(event) => {
-                  // Middle mouse button can also fire onClose
-                  if (event.button === 1) {
-                    event.preventDefault();
-                    onClose(dataKey);
-                  }
-                }}
+              {Icon && <Box marginRight={1}>{<Icon />}</Box>}
+              {/* TODO: add extra info to tooltip from data channel info */}
+              <Tooltip
```
sam-glendenning, line 133: What's your thoughts on making this an arrow tooltip? Might make it easier for the user to see which column is being referred to

louise-davies: I have 0 opinions, I just went with a regular tooltip as the eCat tooltips didn't have arrows

src/table/headerRenderers/dataHeader.component.test.tsx
louise-davies: I'm trying to emulate hovering over the header name specifically, but I suppose we can target the overall sort label as well

src/table/headerRenderers/dataHeader.component.tsx
sam-glendenning: Then I leave this up to the pull request author! I'm all for creative freedom ;)

src/table/headerRenderers/dataHeader.component.test.tsx
louise-davies: It might help if I changed the header name to be more specific so that the label is unique sounding when I target it

sam-glendenning: Ah, because the tooltip is wrapped around the sort label, which is wrapped around the header name, so hovering over the sort icon itself also causes the tooltip to appear. So I guess to hover over the header name itself, we'd have to put the tooltip solely around the header name. Is it worth doing? Unsure. The test does pass anyway

sam-glendenning (approved):
I'll happily give approval for what's here, it does the job fine :)

PR #8, opened 2022-07-06
DSEGOG-15 adding support for icons for columns

sam-glendenning:
## Description
I've added a small range of icons for columns based on mapping the column header name to a specific Material icon. Due to the limited test data we currently have, as well as the huge range of columns we can have, it's hard to define icons for more than the ID, shot number and timestamp columns, which we will always receive data on so can be confident it will appear in the display.

For any remaining columns, I currently show a default `FeedIcon` as it's vague enough to do the job for now. The mapping approach is flexible enough that icons can be easily added or changed later, it just relies on a mapping existing for the `dataKey` of the specified column.

## Column Icons
ID: 
image

Shot Number:
image

Timestamp:
image

Default icon (FeedIcon):
image

## Testing instructions
Note the icons used for ID, shot number, timestamp and channel columns and give thoughts.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-15


src/table/headerRenderers/dataHeader.component.tsx
```diff
@@ -81,14 +83,21 @@ const DataHeader = (props: DataHeaderProps): React.ReactElement => {
       direction={currSortDirection}
       onClick={() => onSort(dataKey, nextSortDirection)}
     >
-      <Typography noWrap sx={{ fontSize: 'inherit', lineHeight: 'inherit' }}>
+      {columnIconMappings.get(dataKey.toUpperCase()) ?? <FeedIcon />}
```
louise-davies: If we're defining the Icons like this (aka via a global map & lookup via `dataKey`) - then isn't the `icon` prop obsolete? If so we should remove the code relating to the `icon` prop method.

sam-glendenning: That was my bad for missing that... see my latest commit, which uses the correct method. Keep an eye on snapshot changes

louise-davies (approved):
Looks good - we'll just have to revisit this later once we know the source of Icon info (unless just using the `Feed` icon is fine)

PR #10, opened 2022-07-11
DSEGOG-17 adding Redux state management

sam-glendenning:
## Description
To allow for complex state management, I am adding Redux. I have based a lot of what I've done on how DataGateway implements Redux, intending to correctly segregate SciGateway state and OperationsGateway state. This also allows for us to start making use of an OperationsGateway settings file, and I've done so by adding in an apiUrl, routes array and pluginHost (none of which are really being used at the moment but will be later on). 

It's likely that, in my attempt to mimic Redux in DataGateway's monorepo structure, I have added something unnecessary or missed something out that we need. Louise will know more about this and can help here.

## Testing instructions
Look over the code and see how things look. I tested that an apiUrl was being read from the settings file by console.logging it in the useRecordsPaginated function, where it is fetched for future use. I've done my best to add tests that ensure the correct actions are firing to store info in state but always good to check this yourself.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-17


src/index.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+jest.mock('loglevel');
+jest.mock('axios', () => ({
+  __esModule: true,
+  ...jest.requireActual('axios'),
+  get: jest.fn().mockResolvedValue({}),
+}));
+
+// TODO test index.tsx (currently struggling to mock axios)
+describe.skip('index - fetchSettings', () => {
```
sam-glendenning, line 15: Would love some help on this one - been struggling to test index.tsx as I'm struggling to mock axios. With this current setup, we get an error of
```
TypeError: _axios.default.get.mockResolvedValue is not a function
```
This makes it impossible to test index.tsx at the moment. Any ideas on how to get this working?

sam-glendenning:
Improving tests on `App.tsx`

src/index.test.tsx
louise-davies: With axios, I always found it better to use a mock in the `__mocks__` folder as we never actually want original axios functionality in the tests at all, so it should always be mocked by default

louise-davies:
@sam-glendenning I think it's safe to close this PR in favour of the redux-toolkit one?

sam-glendenning:
Yes, I was gonna leave it until the other was merged but it was always gonna happen anyway.

Closing in favour of https://github.com/ral-facilities/operationsgateway/pull/44

PR #13, opened 2022-07-12
DSEGOG-10 sticky timestamp column, sticky column headers, date-time selectors

sam-glendenning:
## Description
_Following [updated information](https://stfc.atlassian.net/browse/DSEGOG-10?focusedCommentId=24900) in the original issue, as well as [me managing to totally bork my original PR](https://github.com/ral-facilities/operationsgateway/pull/6#issuecomment-1181914622) for this issue, I decided to make a new PR instead._ 

This PR does a lot of things :

- made column headers sticky, such that when the table is scrolled vertically, they remain in the viewport. I've simulated this by increasing the number of results per page and setting a maximum height on the table container
- made the timestamp column sticky, such that when the table is scrolled horizontally, the entire column and header remains in the viewport. You can simulate this by adding more columns than can fit in the viewport together
- added a date/time filter option above the table, separate from the component itself. This contains a start and end date filter. Whenever one is either successfully filled with a valid date or cleared, the onChange method is fired, which updates the record query
- made the timestamp column visible at all times, with no option to remove it as per the issue details
- made the timestamp column non-draggable since it always needs to be on the left of the table
- recreated a lot of tests based on the above timestamp column changes
- added some more Cypress tests and generally tried to bump the global coverage in unit tests. _Check snapshots carefully_ as there have been many changes made and we want to make sure we get the expected results

I also noticed a bug on Friday afternoon that was present before starting this issue. When a column is added or removed, the current widths of all columns are put back to their default, erasing all user customised table sizing. I have not issued a fix for this in this PR as it's late on Friday and I wanted to get this reviewed by the time I get from leave. Either we fix it here, if it's easily doable, or we create an issue for it and attack it later.

## Testing instructions
I've tried my best to make sure everything plays together via unit tests but the amount of features in this table is quickly stacking up and will continue to lead to more unintended consequences (see above note about column resizing bug). So please test new functionality with as much existing functionality as possible.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-10


package.json
```diff
@@ -3,17 +3,21 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@babel/runtime": "^7.18.6",
```
louise-davies: Renovate will have to pin these dependencies - it might be easier to just pin them here so renovate doesn't get angry 😆

src/app.types.tsx
```diff
@@ -73,20 +68,23 @@ export interface QueryParams {
...
 
   [channel: string]: any;
 }
 
+export interface DateRange {
+  fromDate?: string;
+  toDate?: string;
+}
 // Update this whenever we have a new icon for a specific column
 export const columnIconMappings = new Map()
```
louise-davies, line 88: Can we add some icons for `activeArea` and `activeExperiment`? For `activeExperiment` - `<Science />` seems like a good fit (a conical flask), as for `activeArea` - maybe `Location` (essentially the google maps pin)?

package.json
sam-glendenning: Actually that dependency in particular I don't need, I must have accidentally copied it over so I'll remove it. So I'm a bit unclear, what dependencies need to be pinned and do I do it or Renovate?

src/app.types.tsx
sam-glendenning: Yeah can do! Since we know these columns are more likely to appear on screen, makes sense

package.json
louise-davies: Renovate enforces the practice that we should specify specific version numbers (as Renovate will handle any upgrades) so if remove the caret from any dependencies you install it saves renovate opening a PR to pin them :) It's no big deal if you forget through as renovate will do it otherwise - just saves the extra PR

sam-glendenning: Right gotcha. I'll have to get in the habit of specifying `yarn add -E <package_name>`

louise-davies (approved):

PR #26, opened 2022-07-15
DSEGOG-14 rounding

louise-davies:
## Description
I've added in a fake channels endpoint to represent where we'd get the full channel info from and used that to supply metadata to the columns. We then use a `Cell` renderer on scalar columns to call the `roundNumber` function, which does the rounding according to the spec. Also, since I added all the needed full channel info needed for the tooltips, I modified the tooltips code to do the "show friendly name if available, otherwise show system name" thing.

Also, I noticed some unit tests were throwing console warnings/errors re: the MUI pagination component, so I've either refactored the tests to not throw the errors or just suppressed the error.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check that various test columns have various rounding strategies in the table

## Agile board tracking
Closes DSEGOG-14

### DIFF
--- a/src/app.types.tsx
+++ b/src/app.types.tsx
@@ -24,18 +24,21 @@ export interface Record {
   channels: { [channel: string]: Channel };
 }
 
-export interface Scalar {
+export interface ScalarMetadata {
+  dataType: 'scalar';
   units: string;
 }
 
-export interface Image {
+export interface ImageMetadata {
+  dataType: 'image';
   horizontalPixels: number;
   verticalPixels: number;
   cameraGain: number;
   exposureTime: number;
 }
 
-export interface Waveform {
+export interface WaveformMetadata {
+  dataType: 'waveform';
   xUnits: number[];
   yUnits: number[];
 }
@@ -48,9 +51,22 @@ export interface RecordMetadata {
   activeExperiment?: string;
 }
 
-export interface ChannelMetadata {
-  dataType: Scalar | Image | Waveform;
+export type DataType = 'scalar' | 'image' | 'waveform';
+
+export interface FullChannelMetadata {
+  systemName: string;
+  dataType: DataType;
+  userFriendlyName?: string;
+  description?: string;
+  units?: string;
 }
+
+export interface FullScalarChannelMetadata extends FullChannelMetadata {
+  sf?: number;
+  scientificNotation?: boolean;
+}
+
+export type ChannelMetadata = ScalarMetadata | ImageMetadata | WaveformMetadata;
 
 export interface Channel {
   metadata: ChannelMetadata;
--- a/src/app.types.tsx
+++ b/src/app.types.tsx
@@ -63,7 +63,7 @@ export interface FullCommonChannelMetadata {
 
 export interface FullScalarChannelMetadata extends FullCommonChannelMetadata {
   dataType: 'scalar';
-  sf?: number;
+  significantFigures?: number;
   scientificNotation?: boolean;
 }
 
--- a/src/app.types.tsx
+++ b/src/app.types.tsx
@@ -53,7 +53,7 @@ export interface RecordMetadata {
 
 export type DataType = 'scalar' | 'image' | 'waveform';
 
-export interface FullChannelMetadata {
+export interface FullCommonChannelMetadata {
   systemName: string;
   dataType: DataType;
   userFriendlyName?: string;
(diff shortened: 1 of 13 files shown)


src/app.types.tsx
```diff
@@ -48,10 +51,37 @@ export interface RecordMetadata {
...
+  systemName: string;
+  dataType: DataType;
+  userFriendlyName?: string;
+  description?: string;
+  units?: string;
+}
+
+export interface FullScalarChannelMetadata extends FullCommonChannelMetadata {
+  dataType: 'scalar';
+  sf?: number;
```
sam-glendenning: Honestly this took me a minute to get but now I realise it denotes significant figures. Could we call it `significantFigures` for better clarity?

louise-davies: ah yeah - I didn't think about that - I was too involved in rounding algorithms that `sf` made total sense to me - `significantFigures` is much better. `sigFigs` is another shorthand used in the requirement but I think it's not too long that spelling it out is horrendous

src/table/cellRenderers/cellContentRenderers.tsx
```diff
@@ -0,0 +1,26 @@
...
+): string => {
+  /*
+  In normal mode, do not round to the left of the decimal point,
+  even if the number of significant figures says that you should.
+  For example, 916.3 with a specified precision of 2 sig figs should be
+  displayed as 9.2e2 or 916 in scientific or normal mode respectively.
+  */
+
+  // TODO: what to return here?
+  if (sf <= 0) return 'Invalid significant figure';
```
sam-glendenning: Personally not a fan. As long as we have data to display, I say we do whatever we can to display it. If the rules by which they want us to format that data aren't valid (in this case, an invalid number of significant figures), just return it back to them as is. Therefore, if the input is 916.3, return 916.3. Just my two cents.

louise-davies: Do you mean if we get 916.3 and the significant figures is invalid we return 916.3 (or 9.163e2)? It's hard for us to return 916 or 9.2e2 as those both have implied what the significant figures are (1/2/3 or 2 respectively)

sam-glendenning: Maybe I'm misunderstanding. What sort of scenario would lead to significant figures being <= 0?

louise-davies: I wouldn't really expect it to happen (if the metadata is not junk!) - I just put in the condition as otherwise the function throws an exception if it gets something <= 0

sam-glendenning: Right that's what I thought. What I'm trying to say is that, if significant figures is <= 0, the function can't process it, it's invalid. So just return its input.

louise-davies: yeah that's an easy change :) I think I misread the github notification and read the part of the code comment saying `9.2e2 or 916` as being part of your comment 😅

sam-glendenning: Damn it, that was my bad! I realised I'd mistyped what I wanted to say so quickly tried to edit it but you beat me to it and saw the original comment. Hope my meaning is now clear :)

src/views/recordTable.component.tsx
```diff
@@ -51,15 +55,30 @@ const RecordTable = React.memo(
 
         for (let i = 0; i < keys.length; i++) {
           if (!accessors.has(keys[i])) {
+            const channelInfo = channels?.find(
+              (channel) => channel.systemName === keys[i]
+            );
```
sam-glendenning, line 60: So just to confirm that I have this all in my head correctly:

1. `useChannels` fetches all the info and metadata about every possible channel that could theoretically be returned to the frontend
2. Lines 58-60 are about matching up a column to a possible channel (with a set of metadata) describing it. Records maintain a list of channel names so we take each channel name and match it with its metadata
3. The matching channel info is added to the column object
4. Once all columns are constructed correctly, they are stored and passed to the Table component

One thing I do want to ask is (assuming I have all this right) is where are we getting the info about the channels in step 1? Will there be a new API endpoint for this?

louise-davies: We don't know _exactly_ where we'll be getting the channel info from (either via the og-api or via an external one) but it's definitely not going to be via the records endpoint. E.g. there's going to be a channel selector where the user can select any of the potential channels, Kevin has stated there isn't full channel metadata returned via the records endpoint, records only have a subset of all the potential channels etc.

The exact way of columns accessing the data channel metadata I'm still kind of unsure of (see the later Redux PR) as there's two kinds of column info - the "static" info that comes from the server and then the "dynamic" info like the user's word wrap setting for that column. Either way - this way works well enough for this PR and more accurately reflects what the final workflow will be (aka fetching channels separately from records) and doesn't overhaul the existing code too much.

sam-glendenning (dismissed):
After testing this with https://github.com/ral-facilities/operationsgateway/pull/27, I'm happy for this to be merged

sam-glendenning (approved):

PR #28, opened 2022-07-15
DSEGOG-11 word wrapping columns

louise-davies:
## Description
Adding word wrapping to column headers

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-11


sam-glendenning:
Found a small bug, whereby word wrap on one column is reverted back to off if another column is added or removed

https://user-images.githubusercontent.com/71134574/181044307-90d3daa8-1ad2-45a7-93dc-14f063ac46ff.mp4

So this seems to be another consequence of the column widths reverting back to default as per https://stfc.atlassian.net/browse/DSEGOG-39?atlOrigin=eyJpIjoiNTQxN2E3NDkzOWFhNDhmOWE0YzFiYWMxMGNiY2M4ZmQiLCJwIjoiaiJ9. However, this problem may not occur in your Redux branch since you believe you've fixed the column widths bug (although I've yet to check this myself). If so, I'm happy to let this bug get merged into develop and then we update the redux branch from develop to make sure the bug is properly fixed there

sam-glendenning:
Ugh, broken video. Give me a second to try again

louise-davies:
yeah, this is the same bug that I would expect to be fixed by my redux PR - either by the changes I've already made or via storing the `wordWrap` state in redux itself

sam-glendenning:
Here's what the video should have been.

attempt 2

I'll double check to make sure the bug is properly fixed when I review the other branch so I'll put a pin in this for now

louise-davies:
@sam-glendenning there's a JIRA ticket tracking the resizing reset issue right? I'll add that the same thing happens to word wrapping, and then we can either close that issue if the redux PR fixes things or tackle it later

src/table/columnCheckboxes.component.tsx
```diff
@@ -26,11 +26,10 @@ const ColumnCheckboxes = (props: ColumnCheckboxesProps): React.ReactElement => {
   };
 
   const checkboxes = availableColumns.map((column: Column) => {
-    const header = column.Header?.toString();
     const accessor = column.accessor?.toString();
-    return header && accessor && accessor.toUpperCase() !== 'TIMESTAMP' ? (
+    return accessor && accessor.toUpperCase() !== 'TIMESTAMP' ? (
       <div key={accessor}>
-        <label htmlFor={accessor}>{header}</label>
+        <label htmlFor={accessor}>{accessor}</label>
```
sam-glendenning: So it looks like this change makes the checkbox labels go back to using the accessor, which I think is more likely to be the `systemName`. Would we rather show the user the `userFriendlyName` since this is the title of the column displayed and differs from what the label on the checkbox is?

sam-glendenning: Alternatively, use the `userFriendlyName` as the accessor?

src/table/table.component.tsx
```diff
@@ -141,9 +138,15 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
     onChecked(column, false);
   };
 
+  const handleColumnWordWrapToggle = (column: string): void => {
+    const col = visibleColumns.find((col) => col.id === column);
+    if (col) col.wordWrap = !col.wordWrap;
+    setColumnOrder(columnOrder);
```
sam-glendenning, line 144: Is this like a way of refreshing the table? What's its purpose?

src/table/columnCheckboxes.component.tsx
louise-davies: I don't think the `userFriendlyName` should be the accessor - as for some this may be undefined and so you have a mix of `userFriendlyName`s and `systemName`s, and accessor is important for `react-table`. I'll use the same logic as the headers to display the text for either `userFriendlyName`, `systemName` or `accessor` in that preference order

src/table/table.component.tsx
louise-davies: Yes, this is to refresh the table, as otherwise setting `col.wordWrap = !col.wordWrap` doesn't affect anything. I can add a comment to make this clear?

src/table/table.component.test.tsx
```diff
@@ -203,14 +208,31 @@ describe('Table', () => {
     expect(onSort).toHaveBeenCalledWith('shotNum', null);
   });
 
+  it("updates columns when a column's word wrap is toggled", () => {
+    createView();
+
+    let menuIcon = screen.getByLabelText('timestamp menu');
+    fireEvent.click(menuIcon);
+
+    expect(screen.getByText('Turn word wrap on')).toBeInTheDocument();
```
sam-glendenning, line 217: Not an issue but I think the docs say that, if you're specifically checking for an element to be present on-screen, you can just write this as
```
screen.getByText('my text');
```
Although if you're expecting the opposite, you need to be explicit so you actually need
```
expect(screen.queryByText('my text').not.toBeInTheDocument();
```

src/table/columnCheckboxes.component.tsx
sam-glendenning: Oh yeah I hadn't thought of that. In that case it's up to you because maybe a mix of `userFriendlyNames` and `systemNames` as labels for checkboxes might be confusing

src/table/table.component.tsx
sam-glendenning: Yeah I thought so, that's quite neat. Nah I got it, thanks!

src/table/columnCheckboxes.component.tsx
louise-davies: the mixture will only match what the column headers have i.e. the checkboxes will display userFriendlyNames if they exist, and fall back to systemNames (or accessors) if they're undefined.

Also, this component is only here for testing really as this will eventually need overhauling to be the tree structure selector anyway 😅

src/views/recordTable.component.tsx
```diff
@@ -59,14 +59,27 @@ const RecordTable = React.memo(
...
+              Header: () => {
+                const headerName = channelInfo
                   ? channelInfo.userFriendlyName
-                  : channelInfo.systemName
-                : keys[i], // Provide an actual header here when we have it
+                    ? channelInfo.userFriendlyName
+                    : channelInfo.systemName
+                  : keys[i];
+                // Provide an actual header here when we have it
+                // TODO: do we need to split on things other than underscore?
```
sam-glendenning, line 69: Couldn't see anything in the requirements or design docs that currently indicates so

src/table/columnCheckboxes.component.tsx
sam-glendenning: Well true!

sam-glendenning (dismissed):
Works like a charm!

src/views/recordTable.component.tsx
louise-davies: I'll keep this TODO here, as we currently don't know what the format of the channel names are

src/table/table.component.test.tsx
louise-davies: I don't know if it's a style thing but I think I prefer the explicit assertation - either way it's good to know you can do it either way

sam-glendenning (approved):
Nice improvement :)

PR #44, opened 2022-07-19
DSEGOG-17 Add redux-toolkit for state management

louise-davies:
## Description
Inspired by #10 - this PR uses the new reccomended setup for redux as laid out in the docs. As well as adding an initial `configSlice`, I've also moved most of the internal state we had so far into redux as it's all things we'll eventually have to store in sessions. I also improved some efficiency things and also changed how we pass columns to react table - rather than passing the columns the user has selected, we now pass all the columns and instead use the native column hiding functionality. I believe this fixes the bug mentioned in #13 - as hiding and showing columns no longer loses the column sizing info. We may want to go even further and put the column sizing info in the store so it can be restored via sessions, but I didn't want to go that far with this refactor 😅 

I also added in the theme provider as it was missing and had TODOs associated with it.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes:
- [DSEGOG-17](https://stfc.atlassian.net/browse/DSEGOG-17?atlOrigin=eyJpIjoiYjMzZjNhMGEzYTgzNGJhNjlkYzU0MjBlMDM5YzY1ZTUiLCJwIjoiaiJ9)
- [DSEGOG-39](https://stfc.atlassian.net/browse/DSEGOG-39?atlOrigin=eyJpIjoiMzQ3MGZkZDllZGY0NGMwZTlkNzQzZDdjZWRjZDIwM2YiLCJwIjoiaiJ9)
- [DSEGOG-41](https://stfc.atlassian.net/browse/DSEGOG-41?atlOrigin=eyJpIjoiM2JkMjgxMTAxNTZiNDkyNzgxYTNlN2M4NzY0YTZjZGIiLCJwIjoiaiJ9)


louise-davies:
One thing we need to figure out is the `columnDef`s bit of my code. We probably don't want to store all the column defs in redux as we'll query for these, but we'll need to store things like `wordWrap` and other per-column config (potentially column sizes?). I've left it as it is here as I have another PR that introduces a channels endpoint. Then we'll get to have some fun figuring out how to merge the server-side column definitions and the client-side column config to pass to react-table...

louise-davies:
I have resolved my previous comment by only storing the client side `columnState` in the Redux, and this is used by `Table` via passing in the relevant `wordWrap` prop to `DataHeader`. I did go down the road of merging the `columnState` into the `availableColumns` definitions, but this was inefficient as it caused the whole table to rerender unnecessarily because the `columns` changed.

src/state/slices/columnsSlice.tsx
```diff
@@ -0,0 +1,160 @@
...
+  );
+}
+
+const selectSelectedIdsIgnoreOrder = createSelector(
+  selectSelectedIds,
+  (selectedIds) => selectedIds,
+  {
+    memoizeOptions: { equalityCheck: arrayEquals },
+  }
+);
```
sam-glendenning: Could you explain a bit more what's going on here? We have this function here which just returns the `selectedIds` (this is just an array of the selected column names, right? Like `['timestamp', 'shotnum', 'activeArea']` that sort of thing?). And we also have `selectSelectedColumns`, which I can see is used in `columnCheckboxes` to filter out those which haven't been ticked.

What's the differences between these two functions and why does the order of the IDs (column names?) matter here? We also have that unit test to check that hidden columns selector ignores the order of `selectedColumnIds`. Why does _that_ matter?

I think I'm just getting confused by all the inclusions of "select". The word "select" itself is conventionally used in Redux to refer to selectors. Then we have our definitions of what columns and stuff have been selected by the user. It's just a headache to get around 😅

src/state/slices/columnsSlice.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { Column } from 'react-table';
+import { DropResult } from 'react-beautiful-dnd';
+import { RootState } from '../store';
+import { ColumnState, Order } from '../../app.types';
+import { resultsPerPage } from '../../recordGeneration';
+
+// Define a type for the slice state
+interface ColumnsState {
+  columnStates: {
+    [id: string]: ColumnState;
```
sam-glendenning: I feel like there's potential for confusion here. `ColumnState` and `ColumnsState`. Could we change the latter to `TableState`? I'm also wary of getting in the way of React Table's native `state` object, in case we need this more down the line in `table.component.tsx` but I still feel this needs a name change

src/state/slices/columnsSlice.tsx
louise-davies: `selectSelectedIdsIgnoreOrder` is needed as the `selectedIds` state in redux technically stores 2 pieces of info: the selected columns and the column order. You can see that when columns get reordered they affect this array instead of a separate `columnOrder` state - this is so we don't have duplicate state.

However, a side effect in having selected columns and column order derived from the same redux state is that in the case of knowing the selected columns (i.e. what gets passed as `hiddenColumns` to react-table/as `selectedColumns` to `columnCheckboxes`, order doesn't matter and it would cause unnecessary rerendering if it changed when the column order changed. Hence why we have a `selectSelectedIdsIgnoreOrder` selector which uses memoization to sort the array to provide a "stable" array to serve as `selectedColumns`/`hiddenColumns`.

The unit test for `selectHiddenColumns` is to test the memoization and to ensure that this is effective. You can probably test this makes things more efficient by removing the `memoizeOptions` bit in `selectSelectedIdsIgnoreOrder`, and then try and reorder some columns. I had some noticeable lag without the memoization.

Yeah, I'm trying to use the convention which names selectors as `selectPortionOfState`, but separately there's the idea of "selected" columns.

I should probably write portions of this as comments around these specific selectors to explain what's going on. I spent a lot of time working on the selectors and optimising things so I think by the time I had everything working I kinda forgot about commenting it 😓

src/state/slices/columnsSlice.tsx
louise-davies: Ah true - the convention is for the `State` interface for a slice to share the same name - hence the name as it's the `Columns` slice. I only named this way after the Epic name - this slice could easily be renamed `TableSlice` and thus `ColumnsState` -> `TableState`, `ColumnsReducer` -> `TableReducer`, `state.columns` -> `state.table` etc.

The alternative is to refactor `columnStates`/`ColumnState` - the idea of this is to store all the client-side info of a column - e.g. the word wrap state, potentially the column sizes. Also, we could also refactor the `sort` state to live in there as well. Or, they could all be separate parts of the redux state, so `wordWrap` refactored to look more like the current `sort` state.

src/state/slices/columnsSlice.tsx
sam-glendenning: Right, that explains everything amazingly! But yes that was going to be my next ask - please put some of this in comments! Because I know for a fact I'd like to refer back to this information at some point and I presume some others who work on this project down the line will too. 

So put a comment above the unit test in question to say you are testing the memoization part of the function and put some comments somewhere about what the functions do and how this achieves efficiency in rendering please :)

src/state/slices/columnsSlice.tsx
sam-glendenning: I like the idea of Column and Table respectively. As for where `sort` can live, I think it's all about where makes sense and is easily accessible. Ultimately, it's a parameter passed with several others to the API query, unlike the word wrap which is a client-side piece of info. So maybe it makes sense for those to be two separate entities - `TableState` for how the table is being sorted as a whole and `ColumnState` for how the user has chosen to manipulate columns client-side

Encapsulation, that's the buzz-word I was thinking of

louise-davies: So what are you proposing we do then?
- Rename `ColumnsSlice` to `TableSlice` (and other associated terms in the slice) and keep `columnState` with it's current name?
- Separate out the `columnState` state into it's own slice called `ColumnsSlice`, and rename the existing slice to `TableSlice` as that's more for data passed to the backend? Where would the `selectedIds` state go then? As that presumably will eventually be used to selectively query for data right?
- A third option I didn't understand 😅

src/state/slices/columnsSlice.tsx
louise-davies: @sam-glendenning please read through the comments I added and see if it makes sense & I haven't missed anything :)

src/state/slices/columnsSlice.tsx
sam-glendenning: Apologies, the first option gets my vote

src/state/slices/searchSlice.tsx
```diff
@@ -0,0 +1,51 @@
...
+  initialState,
+  reducers: {
+    // Use the PayloadAction type to declare the contents of `action.payload`
+    changeDateRange: (
+      state,
+      action: PayloadAction<{ range: keyof DateRange; date?: string }>
+    ) => {
+      const { range, date } = action.payload;
+      state.dateRange[range] = date;
+    },
```
sam-glendenning, line 29: Why isn't this part of the columns reducer?

louise-davies: Because the datePicker component is not part of the table component - it instead is going to be part of the search component which will have other things the user can search on like experiment ID (OG10), shot ID (OG11), and timeframe (OG74). The redux-toolkit docs said a good way to approaching slices is to separate reducers according to the main component which uses them, hence why I separated `dateRange`. Both the state from `TableSlice` and `SearchSlice` will have to be passed to the records endpoint, but I figured it was a nice way to split things to prevent a slice from getting too large.

sam-glendenning: That makes sense and I agree with your last point!

src/setupTests.tsx
```diff
@@ -2,8 +2,58 @@
...
+import { initialState as initialSearchState } from './state/slices/searchSlice';
+import { render } from '@testing-library/react';
+import type { RenderOptions } from '@testing-library/react';
+import { Provider } from 'react-redux';
+
+export let actions: Action[] = [];
+export let resetActions = (): void => {
+  actions = [];
+};
+export const getState = (): RootState => ({
```
sam-glendenning: It's a very minor thing but I might support this being renamed to `getInitialState` or `getInitialRootState`, purely because its current name could imply it will return the state as it currently is when called, when that's not actually the case.

louise-davies: True, this is just copying from what it's called in SG/DG. Alternatively - `getMockState` or `mockGetState`? Although I suppose it's primary usecase is to setup the initial state, so perhaps `getInitialState` works better...

sam-glendenning: Up to you! I have personal preference for `getInitialState`. Apologies for the spam comments on this thread, it started out as I only had one thing to say and then it just kept growing... still actively reviewing though!

louise-davies: No worries - I tend to comment as I go along rather than comment all at once so I can't complain at it being done to me! I'll go with `getInitialState` as I think it works with the intention of what it's for more

src/views/recordTable.component.tsx
```diff
@@ -103,100 +88,63 @@ const RecordTable = React.memo(
...
+    },
+    [dispatch]
+  );
+
+  const handleSort = React.useCallback(
+    (column: string, order: Order | null) => {
+      dispatch(changeSort({ column, order }));
+    },
+    [dispatch]
+  );
```
sam-glendenning, line 57: I personally think these could be moved to table? I don't wanna bloat the file too much but they already have some other event handlers like `handleColumnClose`, `handleColumnWordWrapToggle`, etc.

src/views/recordTable.component.test.tsx
```diff
@@ -110,106 +109,108 @@ describe('Record Table', () => {
...
-    });
+    expect(screen.getByTestId('sort shotNum')).not.toHaveClass('Mui-active');
+  });
 
-    await act(async () => {
-      screen.getByTestId('sort timestamp').click();
-      await flushPromises();
+  // TODO: this test is similar to the one in table - either we keep the redux
+  // state in recordTable and do the test here - or we move the state into table
+  // and do the test there
```
sam-glendenning: This kinda relates to my previous comment. What I'm a fan of is testing the correct methods are called and testing that state is correctly altered. This can be more accurate than testing for things on the UI like `"1-12 of ${data.length}"`

sam-glendenning: So to answer your question, leaning towards moving state into table would be my preference, or at least putting all the handlers relating to manipulating the table that call dispatch into the same place

src/views/recordTable.component.tsx
louise-davies: yeah, I wanted some opinions on whether to have all the redux stuff in `Table` and have `RecordTable` just handle the data querying perhaps or to have `Table` be unconnected and put the redux stuff in `RecordTable` and pass it down as props to `Table`. As you can see it created some weirdness in the testing as some stuff was in one place and some stuff is in another...

sam-glendenning: Yeah it's all about figuring out what makes sense to put where, like when I originally had the checkboxes in with the table component because I thought that made sense at the time but now they've been moved back to recordTable, I don't know what I was thinking! It's much better there. So yeah figuring out where things belong can be a tough job

src/views/recordTable.component.test.tsx
louise-davies: I thought the recommendation from `rtl` is to test the app how the user would, so the user would see that when the pagination is handled correctly that the UI text should change, rather than testing that specific Redux actions are sent & that state is updated, which are implementation details that the user can't see.

src/views/recordTable.component.test.tsx
```diff
@@ -110,106 +109,108 @@ describe('Record Table', () => {
...
-    (useRecordsPaginated as jest.Mock).mockReturnValue({
-      data: data,
-      isLoading: false,
-    });
-    (useRecordCount as jest.Mock).mockReturnValue({
-      data: data.length,
-      isLoading: false,
-    });
-    props.resultsPerPage = 10;
+  it('can sort columns and removes column sort when column is closed', async () => {
```
sam-glendenning, line 130: This is an okay test but I looked around and noticed that we don't seem to have any tests that make sure `state.columns.sort` is accurately manipulated whenever we attempt to sort a column. I'd quite like to see this in `columnsSlice.test.tsx` personally but it can go here if you prefer, as long as it's somewhere. 

For example, the check to make sure `state.columns.selectedColumnIds` is correctly manipulated happens in `columnCheckboxes.test.tsx`, instead of `columnsSlice.test.tsx`. This is okay, as long as that test happens somewhere.

src/views/recordTable.component.test.tsx
sam-glendenning: Perhaps you're right. I just don't want to drift so far that we start testing MUI itself. I like testing units of functionality are doing the correct thing, while relying more on Cypress for what the user sees happening.

Anyway, I guess the main thing I wanna say on this particular comment is that, because there's already a fair bit of testing on pagination elsewhere, don't overdo it here. We know the correct page number is sent to the backend thanks to other unit tests

src/App.tsx
```diff
@@ -14,12 +18,37 @@ const queryClient = new QueryClient({
 });
 
 const App: React.FunctionComponent = () => {
+  const dispatch = useAppDispatch();
+  React.useEffect(() => {
+    dispatch(configureApp());
+  }, [dispatch]);
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [_, forceUpdate] = React.useReducer((x) => x + 1, 0);
```
sam-glendenning, line 30: What is this line doing?

src/views/recordTable.component.test.tsx
louise-davies: Yeah, this is mainly there to test the Redux logic for the pagination. Again, I went off of the Redux guide to testing: https://redux.js.org/usage/writing-tests to only test complex selectors/reducers that are hard to test elsewhere in the App, hence there's not many tests in the slice test suites

src/App.tsx
louise-davies: Ah, I meant to link this doc: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate which explains this line. Since `App` is a functional component we can't use `this.forceUpdate()` like we do in DataGateway (https://github.com/ral-facilities/datagateway/blob/73cd741772675501ad1b0bfde1c6d7626270f50d/packages/datagateway-dataview/src/App.tsx#L145) so this is the React recommended equivalent

src/state/slices/columnsSlice.tsx
sam-glendenning: Comments are good!

src/views/recordTable.component.tsx
louise-davies: I've gone for the other way round - I'm extracting all the redux & react-query stuff to `RecordTable`, so `Table` is "dumb" and just deals with `react-table` and styling. I think this separates things logically

src/table/columnCheckboxes.component.tsx
```diff
@@ -1,55 +1,128 @@
...
+        ? column.channelInfo?.systemName
+        : accessor;
+      return (
+        <>
+          <label htmlFor={accessor}>{label}</label>
+          <Checkbox
+            onChange={(e) => handleColumnChecked(e.target.id, e.target.checked)}
+            id={accessor}
+            value={accessor}
+            checked={shouldBeChecked(accessor ?? '')}
```
sam-glendenning: Loving the changes so far! I'm keeping an eye on what's being done so I can provide a timely review when ready.

Regarding this line, out of curiosity, is it more efficient to instead say
```
checked={accessor ? shouldBeChecked(accessor) : false}
```
so that time isn't spent running through `selectedColumns` looking for an empty string?. That being said, I can't think of a realistic scenario when `accessor` isn't DEFINED.

louise-davies: Yes - this makes more sense to do. I think I just went through and added a bunch of `?? ''` to pass TS checks as that's what we do in most other places 😅

sam-glendenning:
@louise-davies are we happy to say this also closes [DSEGOG-41](https://stfc.atlassian.net/browse/DSEGOG-41?atlOrigin=eyJpIjoiYzY1Njc4M2VjMmNkNDkxM2E3NDg4MTEwMTg0MmM2MzEiLCJwIjoiaiJ9)?

louise-davies:
@sam-glendenning Yes I would say so. I did take a different approach using a Backdrop and circular indicator rather than the linear indicator above the table and I'm happy to take feedback on which we prefer UI wise - but it no longer hides & redraws a lot of the UI like it previously did which was the issue Stephen had with it which prompted creating the issue.

sam-glendenning:
Personally I'm a fan of your method but we can see what CLF ultimately think of it. Still reviewing, nearly done

sam-glendenning (approved):
Everything looks to be in working order. This brings a lot of brilliant changes! We now have a more logical placement for a lot of code and a great place to build off of for adding more data to the store and, ultimately, sessions.

From what I can see, this PR closes:
- [DSEGOG-17](https://stfc.atlassian.net/browse/DSEGOG-17?atlOrigin=eyJpIjoiYjMzZjNhMGEzYTgzNGJhNjlkYzU0MjBlMDM5YzY1ZTUiLCJwIjoiaiJ9) (adding Redux for state management)
- [DSEGOG-39](https://stfc.atlassian.net/browse/DSEGOG-39?atlOrigin=eyJpIjoiMzQ3MGZkZDllZGY0NGMwZTlkNzQzZDdjZWRjZDIwM2YiLCJwIjoiaiJ9) (retain column widths after table amendment)
- [DSEGOG-41](https://stfc.atlassian.net/browse/DSEGOG-41?atlOrigin=eyJpIjoiM2JkMjgxMTAxNTZiNDkyNzgxYTNlN2M4NzY0YTZjZGIiLCJwIjoiaiJ9) (improve table loading with loading icon)

I'll add the above to the PR description

PR #52, opened 2022-08-02
DSEGOG-50 add tabs for Data and Plots

louise-davies:
## Description
See title. This justs adds a very simple tabs component that's lets us switch between the Data tab (the current work done so far) and a to-be-implemented plots tab.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-50

### DIFF
--- a/src/views/viewTabs.component.test.tsx
+++ b/src/views/viewTabs.component.test.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import ViewTabs from './viewTabs.component';
+import { screen } from '@testing-library/react';
+import { renderWithProviders } from '../setupTests';
+import userEvent from '@testing-library/user-event';
+
+describe('View Tabs', () => {
+  const createView = () => {
+    return renderWithProviders(<ViewTabs />);
+  };
+
+  it('lets users switch between tabs', async () => {
+    const user = userEvent.setup();
+    createView();
+
+    // should load Data tab initially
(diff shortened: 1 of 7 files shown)


src/views/viewTabs.component.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+describe('View Tabs', () => {
+  const createView = () => {
+    return renderWithProviders(<ViewTabs />);
+  };
+
+  it('lets users switch between tabs', async () => {
+    const user = userEvent.setup();
+    createView();
+
+    // should load Data tab initially
```
sam-glendenning, line 16: I couldn't find the information I needed on this one but is this the behaviour we are wanting or is the user expecting to be able to refresh the page and see the same tab they were already on? Is the user's current tab loaded from a session? This would obviously mean we'd need to put this into the store if so.

louise-davies: I have not seen a requirement for this and it is not the behaviour of eCAT (which always loads the Data view) so I haven't accounted for it.

@mattjSTFC thoughts?

sam-glendenning: His calendar says A/L. I'm happy to let this get merged since we can't find anything specifically saying this behaviour is incorrect. We can always change it later

sam-glendenning (approved):

PR #54, opened 2022-08-04
DSEGOG-4 connecting API to Data tab

sam-glendenning:
## Description
This allows for test data from the API to be displayed in the record table. Time to populate the `apiUrl` field in your settings file! 

The record data itself is requested from `/records` and the record count is requested from `/records/count`. To allow for the display of channel data, I'm also currently using the `/records` endpoint to gather metadata about each channel, as this currently contains a limited amount of metadata for each received channel. This request for all channel metadata is done when the site is launched, to prevent unnecessary extra requests.

Search parameters are now read from the Redux store. Changing `fromDate`, `toDate`, column `sort`, `resultsPerPage` and `page` number should result in a new query to the backend.

I've also added a basic implementation of the `Preloader` component to wait for the settings file to be read before making any backend requests. I copied it straight from DataGateway but if you're sharp you'll notice it's missing its spinning animation for some reason (not sure why but made an issue to fix it later in **DSEGOG -70**).

Near the end, I noticed a bug that was present on develop that meant the timestamp column was not fully sticky (column header was being overlapped by other column headers' menu and divider icons. Fixed this in aa2b6254244b93b5923c9fccea3f4a608b50748b

## Testing instructions
Add in the `apiUrl` provided by Matt to your settings file. Backend data should display! While searching by parameters and changing page, keep an eye on requests being made and make sure React Query is correctly caching requests.

While testing a search by date, you may notice that the way I implemented the DateTimePickers to accept a date value is pretty borked. Sometimes values are accepted and sometimes they aren't. I've made a separate issue for this in **DSEGOG -69** (I won't link it because Jira will think this PR closes it).

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes [DSEGOG-4](https://stfc.atlassian.net/browse/DSEGOG-4?atlOrigin=eyJpIjoiODA3OGEwMmFkMDhkNDk1MDhkYzJkZTQ5OTNhYjA1MDUiLCJwIjoiaiJ9)


louise-davies:
@sam-glendenning it's missing the spinning animation because at least in MUIv4, you couldn't define animations in JS files - so the animations are defined in `App.css`. I don't know if in MUIv5 you can define the animations using emotion which is worth looking into because otherwise people forget that the progress spinner has it's animations defined externally!

src/api/records.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  describe('useRecordsPaginated', () => {
+    it('sends axios request to fetch records and returns successful response', async () => {
+      (axios.get as jest.Mock).mockResolvedValue({
+        data: mockData,
+      });
+      // params.append('limit', '25');
+      // params.append('skip', '0');
+
+      const { result } = renderHook(() => useRecordsPaginated(), {
+        wrapper: renderWithProvidersForHook(),
```
sam-glendenning: @louise-davies I tried to use your `renderWithProviders` function here but couldn't get the syntax right. What parameters should I pass to it if I don't care about rendering any child elements to the screen?

louise-davies: Ah, I'm not sure if it works without a child component. It's not really designed for being used as a wrapper for hooks - I pretty much stole it from the redux docs 😅 

Looking at the DG tests, I think `wrapper` here expects just the wrapping component, and not a `render`. We could probably extract out the wrapper itself to export for hooks testing, and then have another function with the render for use for component testing.

sam-glendenning: That sounds good to me, I can modify what I've done here

sam-glendenning:
I can fix the `Preloader` component if need be but at this point I'm kinda done with adding new features in this PR and just want to get it merged

src/api/records.test.tsx
sam-glendenning: Had a go at fixing it but clearly the end of the week has come and my brain hasn't hacked it. So I've stuck with renaming the render functions to improve their clarity and we can always come back to it later

cypress/integration/table/table.spec.ts
```diff
@@ -155,4 +170,137 @@ describe('Table Component', () => {
...
+
+    cy.get('[role="columnheader"]')
+      .first()
+      .should(($column) => {
+        const { width } = $column[0].getBoundingClientRect();
+        expect(width).to.be.greaterThan(columnWidth);
+      });
+  });
+
+  describe('should be able to sort by', () => {
```
louise-davies: If we're mocking the data - is there any point doing this sort of e2e test? As it's not actually sorting anything right? We already test that we can sort in the unit tests

cypress/integration/table/table.spec.ts
```diff
@@ -155,4 +170,137 @@ describe('Table Component', () => {
         });
     });
   });
+
+  it.skip('should be able to resize a column', () => {
```
louise-davies: This should be removed in favour of the test on `develop`

cypress/support/util.js
```diff
@@ -54,4 +54,339 @@ export function getDraggableSelector(draggableId) {
     return `[${draggable.id}="${draggableId}"]`;
   }
   return `[${draggable.id}]`;
-}
\ No newline at end of file
+}
+
+export const testRecordResponse = [
```
louise-davies: Could this be more useful to be in it's own file and used as a cypress fixture? https://docs.cypress.io/api/commands/fixture#Accessing-Fixture-Data

public/operationsgateway-settings.example.json
```diff
@@ -1,4 +1,5 @@
 {
+  "facilityName": "Generic",
```
louise-davies: Do we need facility name? This was needed for `ICAT` as a lot of features (especially in TopCAT backend) required knowing the facility name. But if our backend doesn't need it we should remove it

cypress/integration/table/table.spec.ts
sam-glendenning: I guess, I think I just added these for when we do need them, then they're ready to be changed for the appropriate asserts. I can remove or skip if you want

cypress/support/util.js
sam-glendenning: Oh yeah, that's a much better place to put it

public/operationsgateway-settings.example.json
sam-glendenning: Oh yeah that can go

src/api/records.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const yKeys = Object.keys(yRow);
+
+    for (let i = 0; i < xKeys.length; i++) {
+      if (xKeys[i] !== yKeys[i]) return false;
+    }
+
+    if (xRow.timestamp !== yRow.timestamp) return false;
+    if (xRow.shotnum !== yRow.shotnum) return false;
+    if (xRow.activeArea !== yRow.activeArea) return false;
+    if (xRow.activeExperiment !== yRow.activeExperiment) return false;
```
louise-davies, line 32: If we iterate over all the keys - isn't it unnecessary to check if the 4 values at the end are the same?

src/api/records.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+      expect(axios.get).toHaveBeenCalledWith(
+        '/records',
+        expect.objectContaining({ params })
+      );
+      expect(
+        dataResponsesEqual(result.current.data, testRecordRows)
+      ).toBeTruthy();
+    });
+
+    it.skip('sends axios request to fetch records and throws an appropriate error on failure', async () => {
```
louise-davies: Why is this test skipped?

src/api/channels.tsx
```diff
@@ -1,29 +1,52 @@
...
+        };
+        metadata.push(newMetadata);
+      }
+    });
+  });
+
+  return metadata;
 };
 
-// TODO change this when we have an API to query
```
louise-davies, line 13: Can we add back this `TODO` to remind us that this is still a WIP way to fetch channels?

src/api/records.test.tsx
sam-glendenning: Truth be told, I didn't think it was worth it atm due to errors simply console logging "Got error". I thought to leave it until we have a proper error handling system, similar to `handleICATError()`

louise-davies: Can we maybe use `todo` then instead of `skip`? `skip` makes me think the test isn't working

cypress/integration/table/table.spec.ts
louise-davies: Eh, if they work they can be left in - I was just thinking if we're not using the API in the e2e tests then these are pointless 😅

src/api/records.test.tsx
sam-glendenning: So basically, line 26 is iterating to make sure the key names of both objects match.

Lines 29-32 check the _content_ of those fields match. But I've been unable to check the content of the channels in each record row match. I think it's to do with additional fields being added in the response or something happens during the base64 encoding.

This is the whole reason I even made this function instead of just using
```
expect(JSON.stringify(result)).toEqual(JSON.stringify(expected));
```
The above returns an error that the two strings parse to the same thing or something like that (but test fails)

cypress/integration/table/table.spec.ts
sam-glendenning: I might as well skip them then, rather not waste time running through them

src/api/records.test.tsx
sam-glendenning: Yep, can do

src/api/records.test.tsx
louise-davies: If you get the `serializes to the same string` error - can't you solve that by using `toEqual`/`toStrictEqual` or something? Normally that error is caused by using `toBe` in the wrong place. Either way - if this is the only way it works then that's alright - I was just a little confused

sam-glendenning: I did try quite hard to use the right comparison (plus plenty of googling) but didn't seem to budge. Since I haven't touched this in a bit, I'll have another go at it again soon before I resolve this convo just to make sure I'm not missing something, because that would be highly preferable to my solution here

louise-davies (dismissed):
I've tested it and it all seems to be working well (pagination, sorting, date filters). I also noticed that datepicker issue - I think it's because the datepicker only commits the change once you've clicked through on everything (even the seconds) - whereas the text in the box updates immediately so it seems like it should search but you're half way through doing the datepicker inputs. However, you've created a separate issue to track this so I'm otherwise happy for this to go in :)

src/api/records.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+      await waitFor(() => {
+        expect(result.current.isSuccess).toBeTruthy();
+      });
+
+      expect(axios.get).toHaveBeenCalledWith(
+        '/records',
+        expect.objectContaining({ params })
+      );
+      expect(result.current.data).toEqual(testRecordRows);
+    });
```
sam-glendenning: @louise-davies I know you'll be amazed this still isn't merged but I thought I'd have one more go at things to try and get the test perfect here.

I've got it working and we can now do direct `expect(x).toEqual(y)` comparisons on this line but at a cost. You'll notice in the select statement of `useRecordsPaginated` that when we have an image or waveform channel, instead of setting the channel data to a raw `` element, I now set it to its stringified equivalent. The raw `` element was introducing difficulties that prevented an easy comparison to an expected result during tests. Now I can just compare strings here.

This means we have to now detect the data type of the channel in `dataCell.component.tsx` in order to render it properly. If the data type is an image or waveform, parse the stringified `` with a third-party HTML parser and render it. Otherwise, just display the passed-in `rowData`. Because of the changes to how `rowData` is passed to the `dataCell`, I also had to change where scalar number rounding happened. This code has also been moved to `dataCell`.

Obviously this is a fair bit of upheaval so I'd appreciate again if you could review the new work. If you think this is not worth the effort, I'm happy for us to roll back to what was previously approved. Either way, I'd like this PR finally merged ASAP so when you see this comment, either approve and merge the new work or roll back to previous approval and merge that!

sam-glendenning:
While looking at other PRs, I found this - https://overreacted.io/why-do-react-elements-have-typeof-property/. Relevant?

src/api/records.test.tsx
sam-glendenning: Following discussion, it was decided to remove this method as its not worth the confusion and goes against React Table methodology. I have now reverted back to commit https://github.com/ral-facilities/operationsgateway/pull/54/commits/40212ac76ae29bca05b20b935c80823da3c75abd

sam-glendenning:
This is ready for a final review, I've effectively reverted back to what we had at https://github.com/ral-facilities/operationsgateway/pull/54/commits/40212ac76ae29bca05b20b935c80823da3c75abd plus whatever else is on develop. @louise-davies please run and check this to make sure everything's good before approving. Also check to see things are working with the plot of time versus shot number. The time axis could maybe be improved by making reference to tips at https://www.chartjs.org/docs/latest/axes/cartesian/time.html but I can do this in my channel plotting PR

src/plotting/plot.component.tsx
```diff
@@ -82,9 +82,10 @@ const ConnectedPlot = (props: PlotProps) => {
   const chartData: ChartData<'scatter'> = React.useMemo(() => {
     const data =
       records?.map((record) => ({
-        x: parseInt(record.metadata.timestamp),
-        y: record.metadata.shotNum ?? NaN,
+        x: Math.floor(new Date(record.metadata.timestamp).getTime()),
```
louise-davies: Very minor comment - what is `Math.floor` doing? I thought `getTime` returns an int...

sam-glendenning: Ah this is leftover from when I was accidentally getting the number of milliseconds since epoch rather than seconds. It's not necessary so I'll change

louise-davies: `getTime` does return milliseconds since epoch - it just returns it as in `number` though instead of a `float`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime

sam-glendenning: Apologies, what I meant was this is from an old calculation. Previously, I had a number that was 1000 times too big, so I was factoring in a division of 1000 that is no longer included in the code

louise-davies (changes requested):
If you filter the data from the current date, the table displays the loading indicator indefinitely instead of displaying no results (which is the result from the API). There's probably some incorrect null checking somewhere...

Can this be fixed so that results of data: `[]` & count: `0` don't trigger the loading logic?

Otherwise this is all looking good!

sam-glendenning:
@louise-davies what do you think about the latest changes? I basically added verbatim the no results message from DataGateway. It appears when we've loaded all data and have a count of zero

louise-davies (approved):
Yep this looks good - we can always change what is shown based on feedback from CLF but at least we're handling it better than showing an infinite loading spinner

PR #60, opened 2022-08-09
DSEGOG-54 plots in windows

louise-davies:
## Description
Building upon the plotting PR, plots are now located in popup windows using React Portals. This means they are still in the same React DOM tree as the main window - which means they share the same redux store etc. so we don't have to do any sort of message passing. I also started the work on the plot grid as shown in the mockups.

I've tested that it works nicely in SciGateway - see gif:
OG-plot-windows-in-SG

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-54


sam-glendenning:
My only thought so far is the names of the files/components - perhaps a little confusing. We have plotGrid, plotWindow, windowPortal, etc. What do you think about plotSelection (since the primary function of this is to select a plot, whereas grid is more its stylistic look), plotWindow can stay the same and plotWindowPortal (since that describes better where and how it's used)?

louise-davies:
> My only thought so far is the names of the files/components - perhaps a little confusing. We have plotGrid, plotWindow, windowPortal, etc. What do you think about plotSelection (since the primary function of this is to select a plot, whereas grid is more its stylistic look), plotWindow can stay the same and plotWindowPortal (since that describes better where and how it's used)?

Yeah I'm not too set on the names of anything - just what first came to mind! `plotWindowPortal` sounds good, `plotSelection ` idk about as you can do other things in it like create & delete plots. Could go for `plotList` to use a more human term than `Grid`? I'm not too set on anything really.

Additionally - I don't know if it's worth it/makes sense to move `plotGrid` & `openPlots` into `views` as that's where `recordTable` is...

sam-glendenning:
Quite possibly, yes! Maybe separate them like views/columns and views/plots ? Up to you

src/state/slices/plotSlice.test.tsx
```diff
@@ -0,0 +1,39 @@
+import PlotReducer, { initialState, createPlot, closePlot } from './plotSlice';
+
+describe('plotSlice', () => {
+  // only test the hard to test bits of columnSlice here - like column reordering logic
```
sam-glendenning: Probably worth removing this to save confusion down the line

louise-davies: oops, my copy-pasting has been exposed 🙈

.yarnrc.yml
```diff
@@ -1 +1,6 @@
 yarnPath: .yarn/releases/yarn-3.2.1.cjs
+packageExtensions:
+  chalk@5.0.1:
+    dependencies:
+      '#ansi-styles': npm:ansi-styles@6.1.0
+      '#supports-color': npm:supports-color@9.2.2
```
sam-glendenning: What do we plan to use this for? Logging?

sam-glendenning:
So from what I can tell this adds the support to rename windows (and plot titles) from main browser to plot windows but not the reverse of this. What issue does this falls under? I presume it will happen when we add other functionality to save the plot exactly as the user left it, with all those fields they edited

.yarnrc.yml
louise-davies: If we don't have this we can't run `serve` -  which I was using to test in scigateway

louise-davies:
> So from what I can tell this adds the support to rename windows (and plot titles) from main browser to plot windows but not the reverse of this. What issue does this falls under? I presume it will happen when we add other functionality to save the plot exactly as the user left it, with all those fields they edited

Yeah I realise there's stuff to do re: storing the plot configs in redux but it was out of scope of this issue. Im not sure there's a specific requirement to save plots (it must come under sessions) - so either we can add an issue to store all plot config in redux, or utilise the "plot can be configured" issue

sam-glendenning:
Yeah I think saving the information isn't worth tackling until it's actually recorded permanently somewhere (i.e. sessions). I'd make sure this is definitely recorded in work to do as part of the sessions epic and leave it until then

sam-glendenning:
I've had an occasional issue where maximising the plot window doesn't work and results in an empty space where the canvas should be. But only in Chrome and Edge and only on the monitor where the rest of the website is? I know that's very specific but can you just check things your end and make sure this isn't too big a problem since some users will likely want to maximise the open windows

louise-davies:
> I've had an occasional issue where maximising the plot window doesn't work and results in an empty space where the canvas should be. But only in Chrome and Edge and only on the monitor where the rest of the website is? I know that's very specific but can you just check things your end and make sure this isn't too big a problem since some users will likely want to maximise the open windows

In firefox, if you maximize and completely cover the original window it still worked for me but the hover behaviour of some of the points on the chart was a little weird. It seems to be the case of this bug in `chart.js` which is under investigation: https://github.com/chartjs/Chart.js/issues/10562. It wasn't too major on FF but I admit I didn't test in other browsers. The only alternative would either be to put an issue in the tracker and hope it's fixed in chart.js at some point, don't use react portals and go through all the issues of normal windows (i.e. need special SG handling code, can't share redux store etc.), or perhaps use a different charting library (if the issue is chart.js specific).

I think it's all because the `window` element in the new window still refers to the original window (this is a feature of react portals), and so event handlers can be a little weird - especially if you cover up the `window`

sam-glendenning:
Considering the work done so far, I really do not want to move away from chart.js, it's a great library which has met all our needs so far and I don't think this is too big an issue in the meantime. It's worth keeping an eye on the mentioned issue though, I'd love for a fix to come out!

Additional mention - a workaround kind of exists anyway. Either use another monitor for maximising it or just awkwardly resize it yourself to not quite maximised (not great but manageable?)

Final edit: other things in the window are also affected, e.g. updating the plot title sometimes works and doesn't (sometimes the user must click elsewhere for the change to be registered)

louise-davies:
I tried with FF and whilst I can still see the plot when maximised, any state updates happen extremely slowly (this may the the "plot title not changing" thing you're seeing @sam-glendenning - for me it just takes a while to update), whereas in Chrome the plot disappears completely. I'll have a look at this again once I come back from holiday as the plot disappearing completely is very odd, as I can see the `canvas` element is still there...

sam-glendenning:
In that case, I'll leave this as is until you return (or maybe have a look myself once I've completed my current work)

sam-glendenning:
I've had another look at this and haven't been able to solve it but did learn some new information that might help down the line, who knows:

1. I only see this issue when the plot window _obscures_ the main browser window entirely. Therefore, this does not mean that the issue is only present in full screen - it just means it will occur 100% of the time in this case because in full screen it covers everything else on the screen
2. In Firefox, when the issue is present, things work but just at a very slow level. Changes made in the plot settings (changing title, changing plot type) can take upwards of 30 seconds to appear in the plot
3. In Chrome and Edge, the plot settings appear but everything in the canvas is missing and doesn't appear on screen (I waited for 5 minutes, nothing happened)
4. If you inspect element, when the issue occurs in Firefox, the canvas element has a small delay where it doesn't have all its style elements. During this time, the plot is not on screen at all. Then when the plot appears, the canvas style is applied in full (during this time, the plot points are all clustered at the bottom of the screen briefly). After a brief moment, the plot points move to their correct position
5. Bullet point 4 does not occur in Chrome and Edge (see bullet point 3)

I had a go at rerendering the plot when a fullscreenchange event fired but couldn't get anything going, [possibly due to this](https://stackoverflow.com/questions/21103478/fullscreenchange-event-not-firing-in-chrome)? I'm at a loss. https://github.com/chartjs/Chart.js/issues/10562 has had no activity since we last checked so unless you can conjure some magic, we might have to leave this one.

louise-davies:
@sam-glendenning I'm still having issues with chart.js even when applying the new PR that was merged via `yarn patch`. I might try out `recharts` for a bit - as they had roughly equiavalent feature sets when I evaluated them, and the main difference is that `recharts` renders as SVGs whereas `chart.js` uses `<canvas>` so that may improve things?

sam-glendenning:
That's a shame, this was looking so promising! And it's so easy to just drag and drop functionality in, like I'm currently doing on the zooming and panning issue, just installed a plugin and away I went.

Give recharts a go because this is a significant issue. Plus if the rendering method is entirely different, we should see different results

louise-davies:
@sam-glendenning `recharts` still had issues in that when you fullscreened, although you could still see the grid none of the points were plotted. I then tried `victory` - which so far has positive results (i.e. the chart resizes normally, allows for mouse interactions i.e. zoom when in a full sized window in chrome) Ignore that it looks ugly as this is just a quick proof of concept

OG-plot-victory-in-window

I'll go through the feature comparison chart I did with the other two libraries to check if `victory` is suitable (I previously dismissed it as it was less popular than the other two and I personally got confused by the docs on first glance, but if it's the only one that supports nice windowing behaviour...)

sam-glendenning:
Interesting, looks like React portals are getting in the way of things for multiple libraries. Let me know where you get with it by the end of the day, since this will affect my work next week

louise-davies:
@sam-glendenning I believe we'll be able to use Victory to fulfil all the requirements - it just requires a little more fiddling/styling than `chart.js` but otherwise, I have full resizing capability, zoom & pan and being able to update the plot config. I might just push what I have now, and leave any fixing of tests until I come back so you can see what I've done/test it out next week. 

GIF for an example - left window is chrome and right window is FF

OG-plot-victory-in-window-2

sam-glendenning:
That looks pretty good! I can work on this next week if you like, I'm nearly done with my channel plotting PR and what you've done here eliminates the work I've been doing on the zooming and panning issue so I'll have some free time

sam-glendenning:
@louise-davies I've managed to do a decent bit of testing on the new changes with the Victory library but have struggled with testing the window resizing events and the `ConnectedPlot` component. This is partly due to the difficulty I've had with properly mocking the Victory components, which has also led to some too-big snapshots with DOM elements that are difficult to parse. You're welcome to see what you can do with this, bearing in mind that I'll be having another go with the `ConnectedPlot` tests in https://github.com/ral-facilities/operationsgateway/pull/62

louise-davies:
@sam-glendenning I've managed to add back in the test for the resize event causing a the plot to redraw - was there anything else you wanted to test differently? You mentioned large snapshots due to being unable to mock Victory, but I can't see any large snapshots...

In any case, the only place we need to mock Victory is when testing `Plot` - everywhere else we can just mock `Plot` itself

sam-glendenning:
Just for reference, I hadn't committed the large snapshots because I knew they wouldn't be wanted anyway. But yeah, we've managed to figure out how to mock things well now so snapshots can be done properly if necessary but most of the time checking the props passed to the mock jest functions is sufficient

sam-glendenning:
I've found that the scale of the plot seems to be a little too big for its container. First, the lowest and highest points on the plot are there but partially obscured because their point lies on where the axes intercept, so the scale is a little too narrow. Secondly, when you test with large numbers on the y-axis, they are partially cut off due to lack of space.

image

Those numbers on the left are actually 366,280, not 66,280. Also, I can't tell if its just because of the size of the numbers involved but changing to logarithmic scale doesn't seem to do anything to the plot itself, although the format of the numbers in the y-axis changes. 

image

Are you getting any of these results or is it just me?

louise-davies:
changing between linear and log definitely works - see my gif above. I think it's just the shotnum scale is so big and the points so close together that changing the scale doesn't visually affect things. See here, where I switch to using another channel (`N_COMP_CALCULATEDE_VALUE`)

OG-plot-victory-scales

As for the other stuff - yeah I'll need to adjust the CSS/sizing a little. Obviously the querying for real data revealed some things!

louise-davies:
I wonder if the points cropping is fine, as this happens in eCAT as well:

image

This can be fixed by adding some `domainPadding` if we want to - thoughts?

sam-glendenning:
It was just a thought, I thought it might be nice to improve on areas that might be lacking. Can you post a screenshot of what it looks like with `domainPadding`?

louise-davies:
Yeah, it's just whether this affects how a user reads the graph, as they might prefer something intersecting an axis to indicate that it's exactly the axis minpoint etc.

This is `domainPadding={size of data point}` (so either 3 for a scatter plot or 2 for a line plot)

image

image

sam-glendenning (approved):
Great work, this was a tough one but you cracked it!

PR #62, opened 2022-08-10
DSEGOG-56 DSEGOG-51 DSEGOG-63 plotting channels

sam-glendenning:
## Description
User can now select fields to plot on both x- and y-axis. This includes a selection of any scalar record metadata, plus any scalar channels that are available. 

I've tried my best to construct the UI based on what the wireframes look like but it's not exactly intuitive. Basically, when a channel to plot is chosen, a label indicating its selection appears under the text input. Currently, only one x- and y- axis can be chosen. In the wireframes, the y-axis tab contains two text inputs/dropdown selectors, plus an Add button, presumably to add the selection to the chart. But which one, because there are two? So to avoid confusion, both work independent of each other and add channels in the same way as the x-axis tab does.

Also because both axes can have data channels on them, I'm going to say this addresses both DSEGOG-56 (plot a data channel on the x-axis) and DSEGOG-51 (able to plot time on x-axis and channel on y-axis).

## Testing instructions
Try plotting channels against channels on the chart (both the dropdown and autocomplete elements should help). The test data doesn't allow for much demonstration but if you plot two channels that are right next to each other in the dropdown list, one point should appear on the plot, since they will both be channels of the same record. Otherwise, you can also plot all the record metadata, which at least have values for each record.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-56 DSEGOG-51 DSEGOG-63


src/plotting/plot.component.tsx
```diff
@@ -70,25 +69,76 @@ export const Plot = (props: FullPlotProps) => {
...
+            break;
+          default:
+            if (Object.keys(record.channels).includes(XAxis)) {
+              const channel = record.channels[XAxis];
+              formattedXAxis = parseInt(channel.data);
+            }
+        }
+
+        // If no valid x value, we have no point to plot
+        if (!formattedXAxis) return { x: NaN, y: NaN };
```
sam-glendenning: I've optimised this a bit but this still results in a lot of null values being passed to the plot. If we can figure out an efficient way of preventing this, without running through the array twice using combinations of map, filter and reduce, that would be handy

sam-glendenning:
This PR also addresses (and should close) DSEGOG-63.

- When time is selected as an axis, the Scale radio buttons are disabled and 'time' is hard locked as the scale in the background until the user removes time from the axis
- Otherwise, user can select only between linear and logarithmic for any other channel

louise-davies:
@sam-glendenning I'm pretty sure the Add button in the wireframes is to add an additional channel to the plot, a la [DSEGOG-52](https://stfc.atlassian.net/jira/software/projects/DSEGOG/boards/22?selectedIssue=DSEGOG-52). See eCat for examples, but the user can both plot multiple channels at once as well as 2 different Y axes (with channels able to be assigned to either Y axis). However, leave that for future PRs - just wanted to clear up some of your confusion based on me reading the PR description!

sam-glendenning:
I figured as much, thanks for that. I will leave that for a future PR

sam-glendenning:
I'm tempted to open this for review but #54 is close to being merged and I feel like it will bring some further changes required in this one so I'll wait for it first

src/plotting/plot.component.tsx
```diff
@@ -9,15 +9,38 @@ import {
...
+
+export const formatTooltipLabel = (
+  label: number,
+  scale: AxisSettings['scale']
+): number | string => {
+  if (scale === 'time') {
+    return format(new Date(label), 'yyyy-MM-dd HH:mm:ss:SSS');
+  }
+  return label;
+};
```
louise-davies, line 29: This is a nice addition - makes the timestamp tooltips a lot more readable 👍

src/plotting/plot.component.tsx
```diff
@@ -96,43 +119,100 @@ export const Plot = (props: FullPlotProps) => {
...
+export const getFormattedAxisData = (
+  record: Record,
+  scalarChannels: FullScalarChannelMetadata[],
+  axisName: string
+): number => {
+  let formattedData = NaN;
+
+  switch (axisName) {
+    case 'timestamp':
+      formattedData = new Date(record.metadata.timestamp).getTime();
```
louise-davies, line 152: Doesn't this convert a `timestamp` - which is already a number representing seconds from epoch, into a JS Date which is then converted into an int representing seconds from epoch (via `getTime`)? Why do this at all?

src/plotting/plot.component.tsx
sam-glendenning: And a necessary one, otherwise they just come out as Unix timestamps

src/plotting/plotSettings.component.test.tsx
```diff
@@ -85,17 +116,50 @@ describe('Plot Settings component', () => {
...
+    createView();
+
+    const radioGroup = screen.getByRole('radiogroup', { name: 'Scale' });
+    const radioButtons = within(radioGroup).getAllByRole('radio');
+    radioButtons.forEach((radioButton) => {
+      expect(radioButton).toBeDisabled();
+    });
+  });
+
+  it('does not let the user change the Y axis scale if time is selected as the Y axis', async () => {
```
louise-davies, line 130: I'm pretty sure the user should only be able to select time on the x axis and not the y axis - @mattjSTFC can you confirm? If so, we'll just need to filter out `timestamp` from the Y axis selection and remove this sort of logic and testing.

src/plotting/plot.component.test.tsx
```diff
@@ -1,5 +1,65 @@
...
+import {
+  testRecords,
+  renderComponentWithProviders,
+  generateRecord,
+  testChannels,
+} from '../setupTests';
+
+const testScalarChannels = testChannels as FullScalarChannelMetadata[];
+
+jest.mock('victory', () => {
```
louise-davies: Now that we're mocking victory at the top - can we remove the `doMock` call in the `Plot` tests?

src/plotting/plotWindow.component.tsx
```diff
@@ -38,6 +50,9 @@ const PlotWindow = (props: PlotWindowProps) => {
     );
   }, [plotTitle, untitledTitle]);
 
+  const { data: records, isLoading: recordsLoading } = useRecords();
```
louise-davies, line 53: Just a thought for the future - we'll likely need to modify `useRecords` to use the values from the `Hours` and/or `Points` fields if they're filled instead of what's defined in the table view. It might be useful to get some clarification from @mattjSTFC on how exactly the users expect to retrieve the relevant data for plots. I think by default it plots based on what's in the table view, but it fetches all the records (so it ignores the pagination parameters).

src/plotting/plot.component.tsx
sam-glendenning: Timestamp is currently in this format:
image

It's capable of handling milliseconds too but at the moment this is the format it's in. I would have thought Unix timestamps would be a better measurement too

src/plotting/plotSettings.component.test.tsx
sam-glendenning: I was not aware of this, would be good for Matt to confirm

src/plotting/plot.component.test.tsx
sam-glendenning: That would make sense but are we still able to verify what props are being passed to mocked components without making use of snapshots? Or we could just use snapshots across the whole file if you prefer

src/plotting/plotWindow.component.tsx
sam-glendenning: Yes that's correct, as we currently don't have exact clarification on this, this was how I implemented it for the time being. We'll need to track it in Jira issues once we have the info we need

src/plotting/plot.component.test.tsx
louise-davies: Ah, yeah the problem is the mock functions cause issues when done in the `jest.mock` due to them being undefined. If this works it works, it's just a little more repetitive. Unless we have a `describe` block that wraps around the whole test suite and do the `doMock` in a `beforeEach` there that's usable by both components' tests?

sam-glendenning: I might be able to get that working, I admit I'm a little unhappy with the current method too

src/plotting/plotSettings.component.test.tsx
mattjSTFC: Yes, I am pretty sure Louise is correct but I am just trying to see if we have anything documented to confirm this.

src/plotting/plot.component.tsx
louise-davies: Oh, I didn't realise that in the metadata they're actual time strings. I guess I was used to our mock data where timestamp was a unix timestamp. No worries then!

src/plotting/plotWindow.component.tsx
louise-davies: Yeah, I'm happy to leave this as it is in this PR and create a separate task to ensure data is being fetched properly once we know how it needs to work.

mattjSTFC: @louise-davies Can you give me more information on what journey requires more detail?

louise-davies: In this wireframe: https://balsamiq.cloud/sq9hezy/p37ch0t/r9154 there's two inputs: `Hours` and `Points`. I presume these are to control data fetching? Looking at eCat, these are a feature of continuous mode which I'm not sure if we're supporting anymore? If so, are these inputs needed?

Otherwise, I presume since there's no controls in the plot config to control how data is fetched - that the data is fetched based on the inputs the user has specified in the table view? e.g. the time range, experiment id etc. (this is how it works in eCat)

src/plotting/plotSettings.component.test.tsx
louise-davies: Looking at the requirements spreadsheet, in OG51 these comments were made on David's test plan

OG51:

> Facility Staff must have the option to plot a data channel on the x-axis (instead of time)

Test plan:

> Select three channels for plotting, and check without specifying it, that the default y axis parameter is time for each channel.

Comments:

> Need to check scatter plot functionality [SD]. I didn't think the requirement was that the y-axis should default to time, only that it's possible for the x-axis  to not be time. [GP]

This seems to indicate that the Y axis is allowed to be time? Maybe we should leave it and when demoing the plotting functionality we can double check then

louise-davies (dismissed):
The functionality is working well and the code looks good - I think the other comments can be saved for later tasks rather than holding up this PR other than the testing one, which feel free to try it out if you think it can work.

src/plotting/plotSettings.component.test.tsx
mattjSTFC: I'll send Stephen and Victoria an email to clarify

src/plotting/plotWindow.component.tsx
mattjSTFC: Thanks for the details. I am not certain either on the definitive purpose of Hours and Points. Let me roll add this question alongside the Time issue and I'll try and get a response back from the CLF team.

src/plotting/plotSettings.component.test.tsx
sam-glendenning: @mattjSTFC it seems likely that this PR will be merged before we get a solid answer from them so when you do get a response either comment it here (I'll still get notified even after merge) or email me and either way I'll make any change they suggest

src/plotting/plot.component.test.tsx
sam-glendenning: @louise-davies would you mind looking over my last commit which fixes this issue? Then I believe I'm happy for this to be merged if you are

src/plotting/plotSettings.component.test.tsx
louise-davies: Or just create a task in JIRA if it needs changing

louise-davies (approved):

PR #73, opened 2022-09-08
DSEGOG-64 Export plots

louise-davies:
## Description
Added buttons that reflect the wireframes and add functionality to export a plot as an image or export the plot data as a CSV.

I rejiggled some things around to make things nicer - i.e. refactored `useRecords` to be generic and thus make `usePlotRecords`, delete `ConnectedPlot` and move it's functionality to `usePlotRecords` etc.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test you can both export a graph as an image (in the form of an SVG) and the graph data as a CSV

## Agile board tracking
Closes DSEGOG-64


src/plotting/plot.component.tsx
```diff
@@ -29,18 +23,27 @@ export const formatTooltipLabel = (
 };
 
 export interface PlotProps {
-  data?: unknown[];
+  data?: Record<string, unknown>[];
```
sam-glendenning: Not related to `Record` in `app.types`?

sam-glendenning: This would also require changing the type of `testData` at L14 in `plot.component.test.tsx`

src/plotting/plotWindowPortal.component.tsx
```diff
@@ -57,6 +57,18 @@ class PlotWindowPortal extends React.PureComponent<
...
+      externalWindow.document.head.appendChild(element);
+
+      const sheet = element.sheet;
+
+      var styles = 'body {';
+      styles += 'margin:0;';
+      styles += '}';
+
+      sheet?.insertRule(styles, 0);
+
```
sam-glendenning: What is this for?

sam-glendenning (changes requested):
SVG exporting works very well and also works when zooming in such that it only outputs what's currently displayed in the plot section zoomed into!

There's a couple of things I might like changed about the plot data exporting:
1. when plotting against timestamps, they are formatted as unix timestamps and not as meaningful date-time strings. It would be useful to convert these using the `formatTooltipLabel` function
2. When the plot contains { x: NaN, y: NaN } points, these are also output to the CSV file. I think it's time we start filtering them out from the data completely. To be honest, the ultimate solution is to do this in `usePlotRecords` so they're never present in the plot to begin with, plus we stop it from potentially being given a large dataset of NaNs that it can't deal with efficiently

src/plotting/plot.component.tsx
louise-davies: No, [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) is a TypeScript keyword. I did have some issues with clashes at some point. I could go back to using `{[key: string]: unknown}[]` as that's essentially what `Record` is

sam-glendenning:
Also we can make the codecov a little happier by adding coverage for L119 and L121 in `app.types.tsx` (I normally wouldn't bother requesting such a small bump in coverage but they're the simplest functions ever so it's trivial and I like seeing ✔️)

EDIT: On second thought, we don't have a test file for app.types and these functions aren't being used anywhere else so no big deal

src/plotting/plotWindowPortal.component.tsx
louise-davies: Most browsers have a set of default styles. One of them is a margin on the `body` element: https://browserdefaultstyles.com/#body. This was messing up the custom styling in the window, so all this code does is set `.body { margin: 0}`. In the main document I presume MUI deals with default styles but we have to deal with it manually in our popups

louise-davies:
@sam-glendenning yeah I only created those functions for completeness and so they can be easily used in the future

src/plotting/plot.component.tsx
sam-glendenning: That's fine, I had to look up what a Typescript `Record` is but I think I've got the gist of it

sam-glendenning (approved):
Nice fixes

PR #75, opened 2022-09-12
DSEGOG-68 add husky & lint-staged to lint files on commit

louise-davies:
## Description
I've had a few commits fail the linting CI check, and this can be avoided by running the linters on commit via lint-staged & husky. I've tested that it lets you commit a valid file without complaining as well as not letting you commit an invalid file.

Also, I've copied the eslintrc file from SG/DG with some minor changes, this applies more linting rules such as cypress/typescript rules.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Try committing files that break the linting rules (easiest is to add an unused import/variable)

## Agile board tracking
Closes DSEGOG-68


src/plotting/plotList.component.tsx
```diff
@@ -38,6 +38,7 @@ const PlotCard = (props: { plotTitle: string; plotConfig: PlotConfig }) => {
   );
 };
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
 interface PlotListProps {}
```
sam-glendenning, line 42: Why do we have this here again?

louise-davies: It's just a standard thing - makes it easier to add new props this way. Although I'm not sure whether we'd ever have to pass props to this, so perhaps I should just delete it

sam-glendenning: You can if you want, it's not a major change or anything, otherwise I'm happy to approve it

sam-glendenning (approved):

src/plotting/plotList.component.tsx
louise-davies: I'll leave it then - we can always remove later if we really know we'll never need to pass props

PR #77, opened 2022-09-13
DSEGOG-52 multiple channels in plot

sam-glendenning:
## Description
This allows for multiple data channels to be plotted in the y-axis. I've removed the <Select> box for choosing channels as it basically does the same thing as the <Autocomplete> but with less functionality so it just gets in the way. This should allow for slightly easier channel selection.

Due to the wide variety in scalar values, plots can look a little weird if totally unrelated channels are plotted side by side. For best results, use two related ones, such as {something}-XPOS and {something}-YPOS.

## Testing instructions
Try plotting multiple channels in the y-axis. Scatter and line graphs should both work and produce nice looking plots.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-52


src/plotting/plot.component.tsx
```diff
@@ -8,39 +8,39 @@ import {
...
-  if (label instanceof Date) {
+export const formatTooltipLabel = (
+  label: number,
+  scale: XAxisSettings['scale']
+): number | string => {
+  if (scale === 'time') {
     return format(label, 'yyyy-MM-dd HH:mm:ss');
   }
   return label;
 };
```
sam-glendenning, line 29: I've gone back to just using numbers only for all labels, including timestamps, hence why you no longer see 
```
{ [channel: string]: number | Date }
```
This just makes it much easier to manage the sorting and ordering of data. This function is called when viewing a VictoryLabel and when converting unix timestamps for export to CSV, which are the only instances unix timestamps turn into human-readable form

louise-davies: No problem, I just thought if we were converting from a `Date` to a unix timestamp and then back to a `Date`again it might be easier to just keep it as `Date`, but yeah it does complicate the types a little. The other thing was it meant I didn't have to match the timestamp column name to it's data in the select function (as you could infer timestamp from the type of `Date`), but you've fixed that :)

src/api/records.tsx
```diff
@@ -194,12 +195,12 @@ export const useRecordsPaginated = (): UseQueryResult<
...
   axisName: string
-): number | Date => {
-  let formattedData: number | Date = NaN;
+): number => {
+  let formattedData = NaN;
 
   switch (axisName) {
     case 'timestamp':
-      formattedData = parseISO(record.metadata.timestamp);
+      formattedData = new Date(record.metadata.timestamp).getTime();
```
louise-davies: Is it not better to keep using the `parseISO` function as that's what `useRecordsPaginated` uses? Just for consistency

sam-glendenning: Okay so I had to look this up to double check exactly what the difference is. As I understand it, `parseISO` takes into account ISO-formatted date-time strings, which is basically what the API is giving us right now. So `parseISO` is the best function here and I'll switch that out to make use of it.

In this instance, we specifically need this date-time value as a unix timestamp, hence why I also call `getTime()` on it. But I presume you're referring specifically to the use of `new Date()` and not this part.

louise-davies: Yeah I was referring to the `new Date` bit 👍

louise-davies (changes requested):
Seems to be working fine - only problem is that when a user adds a channel, they have to clear the channel select box before they can type in a new channel. Could we auto clear this?

louise-davies (approved):
That's a lot better now that we don't have to clear the text box. Also, good catch on the duplicate channels!

Looks good :)

PR #80, opened 2022-09-16
DSEGOG-62 showing/hiding plot channels

sam-glendenning:
## Description
This allows the user to toggle the visibility of a plotted data channel in the plot by toggling a button in its channel label. Note this does not remove the plotted data from the plot - it only toggles its opacity and removes its label from the legend, thus keeping the plot scales the same.

This attempts to replicate the behaviour in eCat. If a channel's visibility is toggled off and we export to SVG, it will export exactly what is shown in the plot, i.e. minus the channel's data. If we export the plot data, all the data in the plot will be exported, including hidden channels. This is the same as in eCat but can be changed if we want to.

## Testing instructions
Use the visibility toggle to show/hide channels in the plot. Try exporting data to test the expected results.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-62

### DIFF
--- a/src/plotting/plot.component.tsx
+++ b/src/plotting/plot.component.tsx
@@ -15,6 +15,7 @@ import {
   PlotDataset,
   PlotType,
   YAxisSettings,
+  SelectedPlotChannel,
 } from '../app.types';
 import { format } from 'date-fns';
 
@@ -30,6 +31,7 @@ export const formatTooltipLabel = (
 
 export interface PlotProps {
   datasets?: PlotDataset[];
+  selectedChannels?: SelectedPlotChannel[];
   title: string;
   type: PlotType;
   XAxisSettings: XAxisSettings;
@@ -39,8 +41,16 @@ export interface PlotProps {
 }
 
 const Plot = (props: PlotProps) => {
-  const { datasets, title, type, XAxisSettings, YAxesSettings, XAxis, svgRef } =
-    props;
+  const {
+    datasets,
+    selectedChannels,
+    title,
+    type,
+    XAxisSettings,
+    YAxesSettings,
+    XAxis,
+    svgRef,
+  } = props;
   const [redraw, setRedraw] = React.useState(false);
   const setRedrawTrue = React.useCallback(() => {
     setRedraw(true);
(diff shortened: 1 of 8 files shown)


src/plotting/plotSettings.component.tsx
```diff
@@ -512,11 +552,33 @@ const PlotSettings = (props: PlotSettingsProps) => {
                     padding: 1,
                   }}
                 >
-                  <Typography noWrap>{plotChannel}</Typography>
-                  <StyledClose
-                    aria-label={`Remove ${plotChannel} axis`}
-                    onClick={() => removePlotChannel(plotChannel)}
-                  />
+                  <Typography noWrap>{plotChannel.name}</Typography>
```
sam-glendenning: This has been a pain. The text does not truncate. Therefore, if you click on a really long channel name, the entirety of the label is rendered and the width of the container element is extended to match, creating a horizontal scroll bar. I've tried to fix this by adding `xs` this and `width: 'inherit'` that but had no luck so any help would be appreciated here for style points.

louise-davies: Honestly, I think the easiest solution here is to set a sensible `maxWidth` on the `Typography` based on the drawer width - everything else. Experimentally, `190px` works well

src/plotting/plot.component.tsx
```diff
@@ -112,47 +122,64 @@ const Plot = (props: PlotProps) => {
...
-                  YAxesSettings.scale
-                );
-                return `(${formattedXLabel}, ${formattedYLabel})`;
-              }}
-              labelComponent={<VictoryTooltip />}
-            />
-          </VictoryGroup>
-        ))}
+        {selectedChannels.map((channel) => {
+          const currentDataset = datasets.find(
```
louise-davies, line 132: I wonder, if we're structuring it that we have the selected channel config in one prop and the data in another - is it not easier/more efficient to have `datasets` just be a single data array, with objects containing fields equalling the channel names? Then, when creating the lines, we go through the list of `selectedChannels`, using the name as the key for the Y axis for each line and the same data array passed as `data`?

i.e. rather than passing `datasets={[{name: "shotnum", data: [{"time": 1, "shotnum":1}, {"time": 2, "shotnum":2}, {"time": 3, "shotnum":3}]}, {name: "activeArea", data: [{"time": 1, "activeArea":1}, {"time": 2, "activeArea":2}, {"time": 3, "activeArea":3}]}]}`, we pass in `data={[{"time": 1, "shotnum":1, "activeArea": 1}, {"time": 2, "shotnum":2, "activeArea": 2}, {"time": 3, "shotnum":3, "activeArea": 3}]}` (we reduce repeated X axis info this way)

Again, this isn't necessarily the best way to do things/worth refactoring but it's just a thought

sam-glendenning: Possibly, I just found it easier to determine what dataset matched with what selectedChannel info by just adding a name field to the dataset, instead of having to extract this from the data itself. What do you think?

src/plotting/plotSettings.component.tsx
sam-glendenning: Guess I was trying to be too clever! `maxWidth="190"` works very well, thank you :)

src/plotting/plot.component.tsx
louise-davies: We can keep it like this for now - we can check later on if things need changing once we do more larger scale tests

louise-davies: Although in the method I propose there is no difficulty in determining the dataset - as there's only one dataset that is supplied to all the lines, and just the y axis value is adjusted

louise-davies (approved):

PR #84, opened 2022-09-23
DSEGOG-59 DSEGOG-61 plot colours

sam-glendenning:
## Description
Joint PR from me and @louise-davies

This allows the user to change the colour of scatter points and lines for a plotted channel on the graph in real time. This uses the [react-colorful](https://www.npmjs.com/package/react-colorful) library's HexColorPicker to let the user choose any colour they want for a channel.

In addition to this, I have also implemented the functionality from eCat which automatically assigns a colour to a channel when it is selected to be plotted. It assigns up to ten specific colours for ten separate channels before assigning a random colour for any channels after this (these specific colours are the first ten from eCat, not including a grey one which wouldn't have looked great on a white background). 

## Testing instructions
Try choosing colours! Also check that colours are automatically assigned correctly (in the order dark green, dark blue, pink, light blue, etc.)

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-59, DSEGOG-61


sam-glendenning:
Updated with ChartJS

src/plotting/plotSettings.component.tsx
```diff
@@ -80,6 +81,94 @@ const StyledTab = styled(Tab)(() => ({
   width: 10,
 }));
 
+/**
+ * Handles the colours currently in use for plotting different channels
+ * Determines which colours we have available based on which have already been selected
+ * Uses a list of 10 pre-selected colours before generating a random colour beyond this
+ */
+export class ColourGenerator {
```
louise-davies, line 89: Wow - I was just going to iterate over the array of 10-ish colours and not care about reusing colours, but this is super cool! Nice solution!

src/plotting/plotSettings.component.tsx
```diff
@@ -80,6 +81,94 @@ const StyledTab = styled(Tab)(() => ({
...
+  selectedColours: string[];
+  remainingColours: string[];
+
+  // List of colours to generate in order (taken from eCat)
```
louise-davies, line 93: I was gonna choose a colour palette from ColorBrewer e.g. https://colorbrewer2.org/#type=qualitative&scheme=Set1&n=9 - I suppose we can't go wrong with picking what eCat had - they're just a tag ugly imo 😆

louise-davies (approved):
The code works really well - moving on the colour picker immediately updates the graph, and the automatic colour selection code is working well as well.

src/plotting/plotSettings.component.tsx
sam-glendenning: My logic was that they can't possibly complain since these are the colours they currently use ;) but I am very open to a separate colour palette if you wanted me to. I haven't used that website before, would you like to recommend one? We'd have to consider what looks best on a plot with a white background (e.g. no grey or very light colours if possible)

louise-davies: Let's just stick with the existing one - it's easy to change it later. The one I linked was the one I was thinking of (9 class set 1).

sam-glendenning: An alternative for 10, in case we come back to this later - this seems to be a literal rainbow so could work quite nicely - https://colorbrewer2.org/#type=diverging&scheme=Spectral&n=10

louise-davies: Yeah, my only thought against it is that the shades are quite similar because it's a "heat map" style colour scheme - which could be confusing

PR #87, opened 2022-09-27
DSEGOG-169 Refactor to use Chart.js for plotting

louise-davies:
## Description
This PR switches the chart library back to Chart.js, as Victory doesn't have proper dual Y axes support. Whereas Chart.js is the only library I've found which both supports dual Y axes (https://www.chartjs.org/docs/latest/samples/line/multi-axis.html) and discontinous time x axis (https://www.youtube.com/watch?v=GC1Th2tpJSo)

This is a bit of a hack, in that all the Chart.js code is now run in the popup window, so we pass along options and data via HTML data-* attributes, which we then watch via a MutationObserver to then update Chart.js. All the existing features of the graph have been hooked up, and I will post a GIF of it working in SciGateway shortly.

EDIT: here's the SG gifs. Note it looks a little weird as for some reason SG still doesn't like Redux in OG, so I've had to comment out any Redux stuff for this demo - but the plotting stuff doesn't really use Redux for much right now other than the list of created plots & for querying. This was something I discovered a while ago, but since SG is still running React 17 I don't think it's worth exploring why Redux is incompatible for now.

OG in SG in FireFox:
OG-chartjs-in-SG-in-FF

OG in SG in Chrome:
OG-chartjs-in-SG-in-Chrome

I still need to update/write tests for this - but we may need to use e2e testing to test this stuff properly (which requires switching away from Cypress to e.g. Puppeteer).

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
~~Should probably make a JIRA ticket!~~
Closes DSEGOG-169


sam-glendenning:
Don't forget this ;)
```
yarn remove victory
```

src/plotting/plot.component.tsx
```diff
@@ -49,138 +38,135 @@ const Plot = (props: PlotProps) => {
...
+            mode: 'xy',
+          },
+        },
+      },
+      scales: {
+        x: {
+          type: 'time',
+          time: {
+            displayFormats: {
+              millisecond: 'HH:mm:ss:SSSS',
```
sam-glendenning: Not sure what this would do if it got here but just in case, this has one 'S' too many. So should be:
```
millisecond: 'HH:mm:ss:SSS',
```

louise-davies:
Yep, don't worry I noticed when refactoring the tests ;)

src/plotting/plotWindowPortal.component.tsx
```diff
@@ -57,6 +45,141 @@ class PlotWindowPortal extends React.PureComponent<
...
+  
+          observer.observe(document.body, {
+            childList: true,
+            subtree: true
+          });
+        });
+      }
+
+      var waitForChartJS = setInterval(function () {
+        if (typeof Chart !== 'undefined' && typeof Hammer !== 'undefined' && typeof ChartZoom !== 'undefined') {
```
sam-glendenning: Do we also need to wait for `date-fns`? Or is this not necessary because we won't be using it to format dates anymore, since this is covered in `displayFormats`?

sam-glendenning (commented):
Reviewed the code up until the unit tests were pushed, all looks good but see the below:

src/plotting/plotWindowPortal.component.tsx
louise-davies: We don't actually import all of `date-fns` - the import we do for `chartjs-adapter-date-fns` includes any date-fns code it uses. I did try and find the variable so we can wait for it, but I don't think it creates one. Either way, the browser should wait for it to load as I set `async` & `defer` to be false on all the libraries being loaded - these checks were from when I didn't realise some browsers set `async = true` if you dynamically add a script but I figured they act as another safety net so it's good to keep the checks in.

I'll double check if there's a variable we can wait for with the date library tomorrow

src/plotting/plot.component.test.tsx
```diff
@@ -1,267 +1,68 @@
...
+    const { rerender, asFragment } = render(<Plot {...props} />);
 
-      render(<Plot {...props} />);
+    props = {
+      ...props,
+      title: 'line plot',
+      type: 'line',
+      XAxisSettings: { scale: 'linear' },
+      YAxesSettings: { scale: 'log' },
+    };
```
sam-glendenning: Can we also check that `props.XAxis` updates on the graph as well?

src/plotting/plotButtons.component.tsx
```diff
@@ -4,26 +4,23 @@ import { PlotDataset } from '../app.types';
 import { formatTooltipLabel } from './plot.component';
```
sam-glendenning: Safe to say this function belongs in this component from now on, since this is the only place it's being used?

src/plotting/plotSettings.component.tsx
```diff
@@ -360,7 +360,7 @@ const PlotSettings = (props: PlotSettingsProps) => {
                     label="Linear"
                   />
                   <FormControlLabel
-                    value="log"
+                    value="logarithmic"
```
sam-glendenning, line 363: If this is changing, then we should also update the AxisSettings interfaces to say 'logarithmic' instead of just 'log' in `app.types.tsx`. Frankly, I'm not even sure how this is working considering `changeXAxisSettings` should only be accepting a value of time, linear or log. Make sure to do this for both `XAxisSettings` and `YAxisSettings`

sam-glendenning (changes requested):

src/plotting/plot.component.test.tsx
louise-davies: Thanks for catching that 👍

src/plotting/plotButtons.component.tsx
louise-davies: True, I'll move it here

src/plotting/plotSettings.component.tsx
louise-davies: I thought I had changed the value in the types, obviously I'm misremembering. Good catch - it should definitely be changed in the types.

It's not caught as `value` is cast to `XAxisSettings['scale']` as otherwise it's just a `string`, and Typescript only catches build time errors not runtime. it's also then not caught in the chart js settings code as we currently don't have that object types. I could install the chart.js packages as dev dependencies purely for type info? Thoughts?

sam-glendenning: Ohhh that makes sense, I wouldn't have thought of that. Yes, if there's a way to import the ChartJS scales and remove any of our own defined ones, I think we should definitely do this. Ensures that everything is correct and improves maintainability

louise-davies: I was more meaning to type the chart.js options objects, but I can see if we can do that as well. I'll ensure that I add chart.js packages as dev dependencies only

sam-glendenning (approved):
Very nice!

PR #88, opened 2022-09-27
DSEGOG-65 DSEGOG-150 use all records in plotting, choose channels from record table

sam-glendenning:
## Description
Addresses two issues:
- Plots can now use all data available from the records endpoint. Previously, it was only using the paginated data from the record table. 
- Users can now choose scalar channels to plot that are currently present in the record table. This is done via a `Select` above the existing `Autocomplete` which is used to choose from any available channels. 

A significant amount of refactoring has been done to harmonise the channel metadata available to all components which want them. The Redux store previously contained the availableColumns, which only contained information about the channel name. This wasn't useful for the plot to determine which table columns contained scalar data that could be plotted. Therefore, this record of availableColumns has been replaced with an array of FullChannelMetadata, which is more useful for the plot and potentially more useful for future functionality involving data channel searching, viewing images and traces, etc. The only place where availableColumns is still used is in the recordTable, which is the only place that realistically needs that information.

The potential downside to this is that the shot number, active area and active experiment columns are now also treated as data channels to accommodate this. This may not make a lot of sense but was easiest when doing this, although we'd have to still treat those "channels" as special cases when the user is wanting to browse or search channels. I'm open to suggestions on whether this is a good idea and if it should be changed.

## Testing instructions
Populate the record table with a mixture of scalar and non-scalar columns and then try and plot them while making sure only the scalar channels are selectable. Also check that any data that is plotted is not from a paginated query, i.e. it contains all data we have on this channel.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-65 DSEGOG-150


src/api/records.tsx
```diff
@@ -109,18 +109,24 @@ export const useRecords = <T extends unknown = Record[]>(
         resultsPerPage: number;
         sort: SortType;
         dateRange: DateRange;
+        plotRecords: boolean;
       }
     ]
-  >
+  >,
+  plotRecords = false
```
louise-davies: Could name this more generically as `paginate` - as that more accurately represents what this variable is controlling

src/api/records.tsx
```diff
@@ -109,18 +109,24 @@ export const useRecords = <T extends unknown = Record[]>(
...
 ): UseQueryResult<T, AxiosError> => {
   const { page, resultsPerPage, sort, dateRange } =
     useAppSelector(selectQueryParams);
   const { apiUrl } = useAppSelector(selectUrls);
 
   return useQuery(
-    ['records', { page, resultsPerPage, sort, dateRange }],
+    ['records', { page, resultsPerPage, sort, dateRange, plotRecords }],
```
louise-davies: Hmm, thinking of this it could have potential negative side-effects. For example, if a user is plotting, then the user switches the page and opens a new plot - the plot will have to refetch the record data as the React query key has changed - even though the pagination is not relevant to the plotting. It may be that we should in fact separate the `useRecordsPaginated` and `useRecords` (not paginated) query functions. Thoughts? If there end up being more complex shared options than `onError` - we could easily extract those common options to share between them.

src/plotting/plotWindow.component.test.tsx
```diff
@@ -6,9 +6,12 @@ import {
   renderComponentWithProviders,
   testChannels,
   testPlotDatasets,
+  // getInitialState,
 } from '../setupTests';
-import { useScalarChannels } from '../api/channels';
+import { useScalarChannels, useChannels } from '../api/channels';
 import { usePlotRecords } from '../api/records';
+// import { PreloadedState } from '@reduxjs/toolkit';
+// import { RootState } from '../state/store';
```
louise-davies: Remove commented out code in this file

src/views/recordTable.component.tsx
```diff
@@ -31,9 +31,12 @@ const RecordTable = React.memo((): React.ReactElement => {
     useAvailableColumns();
 
   const columnStates = useAppSelector(selectColumnStates);
+
+  const { data: allChannels } = useChannels(); // just for use with hiddenColumns, not passed anywhere else
```
louise-davies: I don't like essentially querying for the same data twice (via `useAvailableColumns` and `useChannels`) just for passing into `selectHiddenChannels`. Since all `selectHiddenChannels` is is a `string[]` and we essentially are just comparing "channel id"/`systemName` - is it possible to bring back `selectHiddenColumns` - i.e. the selector takes in `availableColumns` and uses that to filter the `string[]`? We can have two selectors `selectHiddenColumns` & `selectHiddenChannels` if necessary.

src/plotting/plotWindow.component.tsx
```diff
@@ -56,10 +63,20 @@ const PlotWindow = (props: PlotWindowProps) => {
 
   const { data: records, isLoading: recordsLoading } = usePlotRecords(
     XAxis,
-    selectedChannels
+    selectedPlotChannels
   );
   const { data: channels, isLoading: channelsLoading } = useScalarChannels();
 
+  const { data: allChannels } = useChannels(); // just for use with selectedRecordTableChannels, not passed anywhere else
```
louise-davies: Same criticism as in `recordTable` - can we not just pass `channels` (i.e. the scalar channels) to `selectSelectedChannels`, and then we don't need `selectedRecordTableChannels` - we just have `selectedScalarRecordTableChannels`? We never need to know the e.g. image and waveform channels a user has selected in the table in the plotting component

src/table/columnCheckboxes.component.tsx
```diff
@@ -1,29 +1,20 @@
...
   deselectColumn,
 } from '../state/slices/tableSlice';
 import { useVirtualizer } from '@tanstack/react-virtual';
-import { useAvailableColumns } from '../api/channels';
+import { useChannels } from '../api/channels';
+import { FullChannelMetadata } from '../app.types';
 
 const ColumnCheckboxes = React.memo((): React.ReactElement => {
-  const { data: availableColumns } = useAvailableColumns();
+  const { data: channels } = useChannels();
```
louise-davies, line 14: I actually prefer that after the refactor, we can just use `useChannels` here and it makes the code in here a lot simpler

src/plotting/plotSettings.component.tsx
```diff
@@ -373,7 +354,9 @@ const PlotSettings = (props: PlotSettingsProps) => {
                 freeSolo
                 clearOnBlur
                 id="select x axis"
-                options={axisSelectionOptions}
+                options={['timestamp'].concat(
```
louise-davies: Not sure if this would ever be a problem - but this `options` array is new on each rerender so would cause the `Autocomplete` to rerender whenever the settings panel changes at all. Would just need to extract it to a `useMemo` if we're concerned about it

src/views/recordTable.component.tsx
sam-glendenning: So transform selectHiddenChannels back into what it was before where it was returning an array of react-table Columns and have columnCheckboxes use that as before? Happy to do it for ease of use in the code like you say, just want to make sure I have your meaning right

src/plotting/plotWindow.component.tsx
sam-glendenning: Oh yeah, that would work. I think I got muddled up while doing this over what data I had available, where it was going, etc. Plus the additional useQuery calls where they may not be necessary

src/views/recordTable.component.tsx
louise-davies: `selectHiddenColumns` always returned a `string[]` - the difference is what argument it expects. It makes sense to me for it to accept `Columns` as that's what we query for in `recordTable` (via `useAvailableColumns`). Creating extra selectors is pretty easy - if we ever needed a `selectHiddenChannels` selector that accepts `FullChannelMetadata[]` instead we could create an extra selector rather than refactor `selectHiddenColumns`

src/table/columnCheckboxes.component.tsx
sam-glendenning: Okay hold up a sec, see my earlier comment. Do we want the columnCheckboxes to be passed the availableColumns or the channels from recordTable or what exactly?

src/plotting/plotSettings.component.tsx
sam-glendenning: A good point and another area where there's some inconsistency. Timestamp is not being treated as a channel as such but we still need it to appear in the same place as them in some situations, such as here and as a column in the record table. I'm unclear how to unify all this at the moment

src/table/columnCheckboxes.component.tsx
louise-davies: I mean it doesn't matter really as `columnCheckboxes` is a test component that won't end up in production, but it doesn't really matter which one it uses. `useChannels` makes the code a little nicer imo? It's the same underlying query, just that `useAvailableColumns` transforms it into `Columns`. But we don't need `Column` objects in the checkboxes - we just need a list of IDs essentially

sam-glendenning: Right I'm with you, I think I can tackle this all in my head now!

src/plotting/plotSettings.component.tsx
louise-davies: Might make it mentally easier keep track to treat timestamp the same as the rest, and just filter it out when we don't need it (e.g. column checkboxes, Y channel selection)

src/api/records.tsx
sam-glendenning: I hadn't thought of that. So go back to two separate functions for useRecordsPaginated and useRecords (could rename this to useAllRecords or something, I don't know, just to make it clearer)?

louise-davies: yeah, I think that's the best way as otherwise we could end up requerying lots of results in plotting.

sam-glendenning:
Thanks for the fix! I think this is ready for another review

src/api/records.tsx
```diff
@@ -186,7 +255,7 @@ const useRecordsPaginatedOptions = {
     }),
 };
 
-export const useRecordsPaginated = (): UseQueryResult<
+export const useRecordsPaginatedOld = (): UseQueryResult<
```
louise-davies: Is this used anywhere? If not then we can delete it

src/api/records.tsx
```diff
@@ -98,10 +98,38 @@ const fetchRecordCountQuery = (
 
 // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
 export const useRecords = <T extends unknown = Record[]>(
-  options?: UseQueryOptions<
+  options?: UseQueryOptions<Record[], AxiosError, T, [string]>
+): UseQueryResult<T, AxiosError> => {
+  useAppSelector(selectQueryParams);
```
louise-davies, line 103: This theoretically will need search params like `dateRange`, but we can sort that out later as part of the Search epic

src/api/records.tsx
sam-glendenning: Good spot, leftover junk

src/api/records.tsx
sam-glendenning: Yes, I wasn't quite sure how to leave this, whether we should leave it in its generic state for later customisation later if necessary or if the whole thing should be reverted back. I decided to leave it until we had a better understanding. If it ain't broke, don't fix it

src/table/columnCheckboxes.component.tsx
```diff
@@ -1,29 +1,20 @@
...
 
-  const selectableColumns = React.useMemo(
-    () =>
-      availableColumns?.filter((col) => {
-        const accessor = col.accessor?.toString();
-        return accessor && accessor.toUpperCase() !== 'TIMESTAMP';
-      }) ?? [],
-    [availableColumns]
-  );
```
louise-davies, line 23: If timestamp is now treated like a normal channel, it is returned by `useChannels` right? Which means we need code to filter out timestamp again

src/api/records.tsx
louise-davies: Yeah, we can sort out params related to searching e.g. date ranges etc. for the Search epic as the date range is here because you implemented the search boxes UI a little early

src/table/columnCheckboxes.component.tsx
sam-glendenning: You're absolutely right, another thing I must have missed. I'll also add a unit test to check for this in column checkboxes

louise-davies (approved):
This looks good and fixes that we were only fetching e.g. 25 records in the plots

PR #89, opened 2022-09-28
DSEGOG-58 adding button to reset zooming and panning

sam-glendenning:
## Description
This adds a plot button which resets both the zooming and the panning currently set on the plot to whatever its default calculated value was. This is achieved by setting a boolean flag on the ChartJS `canvas` object, which calls `chart.resetZoom` whenever it changes.

## Testing instructions
Hit the button. Is the zooming and panning reset? Great success!

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes DSEGOG-58


sam-glendenning:
This is okay but doesn't test the view reset functionality terribly well. Ideally, I'd like to spy on the `chart.resetZoom` method but I'm not sure if this can be done or not. Any ideas, Louise?

louise-davies:
Anything testing the Chart.js code will likely need to be done in e2e tests. I have some working tests with Playwright on Firefox which include testing zoom/pan behaviour (what we'd do here in zoom/pan in, take a screenshot, reset zoom and expect to be back to "normal")

I'm just debugging some issues on why Chrome isn't loading the chart libraries when running via Playwright before I can create a PR to add e2e tests for chart.js stuff

sam-glendenning:
That's what I thought, thanks! In the meantime, I'll leave this as is and can be reviewed/merged whenever

louise-davies (approved):
Works well :) I'll just have to add tests in the playwright PR

PR #90, opened 2022-09-29
DSEGOG-170 add visual regression tests for plotting via playwright

louise-davies:
## Description
Added in playwright setup and a spec file which tests all the current plotting features we can't test via unit tests.

I've tested these on my machine via WSL so it should be the same environment at Github actions (i.e. Ubuntu) and thus the screenshots should be stable but we'll see. There's also some oddities which I've commented - e.g. Chrome (via chromium( doesn't load the popup libraries unless you close the main window (this is purely a Playwright issue), Safari (via webkit) has some issues with clicking the toggle visibility button.

This adds a github actions workflow (which is based on the default one playwright sets up). I'd expect any local testing to need to be done via Linux - either via WSL like I did or via some other mechanism (Docker, VM etc.) as otherwise the screenshots will be different.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-170


louise-davies:
wow, can't believe they passed first time on CI 🤯 I guess all the effort into making sure they were stable on WSL was worth it!

sam-glendenning:
Have a bit more faith in yourself, we're supposed to be professionals after all!

louise-davies:
The plawright tests should fail now - because the plots should be a different colour in the screenshots

louise-davies:
Yup, they failed as expected. What's cool is that it uploads the test report: https://github.com/ral-facilities/operationsgateway/actions/runs/3158014219 - which you can download and then use the comparison tool to see what was different in the screenshots (here, it's that the plot is green rather than red!)

sam-glendenning:
I really like that, the Playwright report is very straightforward and the screenshots are automatic and have expected/actual/diff. I feel bad for Matt and how long he spent trying to get Cypress screenshots working now...

louise-davies:
The jury is still out on which framework I prefer for "normal" tests - it may be worth trying it out doing "normal" UI interaction stuff (i.e. not in a popup, not using a canvas so need to do visual regression testing etc.) but if we like playwright we can switch to use it for all e2e tests in OG since we need the popup interaction feature

(I still like Cypress, so I wouldn't suggest switching for SG or DG - just that if we stick with Cypress in OG then we have two sets of e2e tests!)

sam-glendenning:
How do I run the tests locally?

louise-davies:
`yarn playwright test` - however the screenshot comparison will only work on linux - you can however run the tests in headed mode on windows to watch it control the browser `yarn playwright test --headed`. You can do it via WSL as long as you set up X11 forwarding (I used `VcXsrv`)

`yarn playwright test --debug` runs the test with the inspector attached, which runs the browser in headed mode and steps through each command.

`yarn playwright test --update-snapshots` to update snapshots

`yarn playwright test --project "chromium"/"firefox"/"webkit"` to target a specific browser

If you're debugging tests, then I'd recommend running `yarn e2e:serve` in a separate terminal so you don't have to keep rebuilding & reserving OG.

You may also need to run `yarn playwright install --with-deps` to install browsers & any other libraries it needs to run

sam-glendenning:
Right so I've come to the conclusion that in order for the tests to pass locally, you have to be running them on the same architecture as the snapshots were made in. For example, yours are in linux and mine are in Windows so the tests failed because it had to write new snapshots with win32 appended instead of linux

louise-davies:
Yes, it's because Playwright forcibly adds the operating system to the name of the snapshot. Since we need them to pass on CI, that means we need to use Linux. I was developing OG in WSL anyway so it wasn't a big deal for me, but yeah you may need to setup WSL just to run playwright. You can still have your files in the windows side, and just run the command in WSL (WSL can access windows files under `/mnt/` - I can help you setup WSL if you'd like?

Alternatively, you can either use Docker or we can hack the screenshot names as detailed here: https://github.com/microsoft/playwright/issues/7575 (although I would be wary of creating screenshots in windows and expecting them to pass on CI)

sam-glendenning:
I agree, I got halfway through setting up WSL a while back but gave up because yarn wasn't working. I'll try again

sam-glendenning (approved):
I'll approve though because the code looks good and CI seems to like it. Plus from what little of the tests I could do locally, they seemed fine too

louise-davies:
Cool - let me know if you struggle to run the tests locally and we can brainstorm and find a way forward :)

PR #92, opened 2022-10-03
DSEGOG-67 additional plot styling options

sam-glendenning:
## Description
This PR adds:
- ability to toggle plot gridlines (for both axes)
- ability to toggle axes labels (currently only displaying a label for x-axis, since we haven't finalised the y-axes and their respective labelling yet)
- ability to add minimum and maximum limits for both x- and y-axes
- ability to change line style between solid, dashed and dotted

The amount of options per data channel means there's now a lot of buttons clogging up the label space. I'm open to adding some sort of popup menu per data channel - the user clicks a button, which opens a popup showing the options for each channel, which they can edit there

## Testing instructions
Try out the above features mixed with each other to make sure they behave independently

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-67


sam-glendenning:
Right I think this is finally ready for review. Snapshots have changed considerably so check them thoroughly before approving them. This includes plot jest snapshots so make doubly sure that the data and options passed to the `canvas` are correct.

Regarding Playwright, I have mostly managed to fully test all functionality but there are a couple of blockers due to tiny nitpicks some browsers have with particular elements. We may be able to fix these later if we discover a solution. I will add a comment on the PR for each one to highlight them

e2e/plotting.spec.ts
```diff
@@ -68,7 +68,7 @@ test('plots a time vs shotnum graph and change the plot colour', async ({
   const chart = await popup.locator('#my-chart');
   const dimensions = await chart.boundingBox();
   expect(
-    await popup.screenshot({
+    await chart.screenshot({
```
sam-glendenning, line 73: By changing this to `chart`, I've managed to more deterministicly capture the plot itself for a screenshot. Without this, some elements of the chart are sometimes missed in the screenshot, such as the axes labels

e2e/plotting.spec.ts
```diff
@@ -293,7 +293,264 @@ test('plots multiple channels on the y axis and displays correct tooltips', asyn
...
+  ).toMatchSnapshot({ maxDiffPixels: 100 });
+});
+
+test('user can hide gridlines and axes labels', async ({
+  page,
+  context,
+  browserName,
+}) => {
+  // can't test this on chrome with the "have to close main window" workaround
+  if (browserName !== 'chromium') {
```
sam-glendenning: The buttons to activate this functionality are the same as the Reset View button. The comment on line 211 seems to indicate that chrome can't properly test these features (at least until we find some way around it). Therefore, I have told chrome to skip this test altogether (or rather pass it without doing anything)

e2e/plotting.spec.ts
```diff
@@ -293,7 +293,264 @@ test('plots multiple channels on the y axis and displays correct tooltips', asyn
...
+  }
+});
+
+test('user can add from and to dates to timestamp on x-axis', async ({
+  page,
+  context,
+  browserName,
+}) => {
+  // can't test on firefox (date-time pickers aren't editable in headless)
+  if (browserName !== 'firefox') {
```
sam-glendenning: What seems to happen, according to the trace, is that Firefox in headless mode adds a `readonly` prop to the DateTimePickers which isn't present in headed mode or in general testing by myself. Although even by passing `readonly: false` as an `inputProp`, Playwright still refused to fill in the field, despite it saying it was visible and editable. Either way, I couldn't get this to work in Firefox headless, although it works perfectly fine in headed

sam-glendenning:
Also worth noting that the screenshots I generate on my end tend to be a few pixels in width different. Why this is, I don't know and hopefully it isn't a concern but at least the tests pass due to the overall difference being < 100 pixels

sam-glendenning:
I have applied Louise's fix to a webkit test from https://github.com/ral-facilities/operationsgateway/pull/88/commits/4f1baaf4f3f0dcd455ddbeebbaf611e249704501 to ensure things also worked fine on this branch - it seems to!

e2e/plotting.spec.ts
louise-davies: Ah no, the issue with chrome is nothing to do with the buttons themselves! The issue is that Chrome (when run by Playwright) only loads Chart.js when the main window is closed - but since React is running in the main window once we close it anything that works on the React side (e.g. the buttons) will no longer work. Since the specific Chrome test we skipped in Chrome was testing zoom (which you need Chart.js loaded for - so we need to close the main window) and then resetting the Zoom (which is done via React and so needs the main window) it can't work on Chrome. There should be nothing wrong with this test for Chrome as you can toggle the grid lines and axes labels before Chart.js actually loads!

sam-glendenning: Gotcha, I'll have another go at it, thanks :)

louise-davies:
@sam-glendenning I agree that the options per data channel are getting a little long - one space saver is that we only need line style options when it's a line graph as the options do nothing for scatter graphs ;) Also, it looks like there's some extra unnecessary padding between the visibility button and the line style button.

@mattjSTFC There's a lot of options that CLF want in the requirements that aren't accounted for in the wireframes. Specifically, there's these "per-channel" options (e.g. line style, colour, visibility, removing a channel) that are starting to get a little cramped (especially since I have to add an extra option to select whether an option is on the left Y axis or on the right Y axis)

image

Any thoughts on a better UI for this? As Sam said - some sort of popup menu? Which is a little weird with the colour selector as that's then a popup within a popup!

e2e/plotting.spec.ts
louise-davies, line 73: Hmm, I didn't know that we could take a screenshot of the element itself - this means we don't have to set the dimensions of the screenshot right? But I suppose if it's not broke then don't fix it!

src/plotting/plot.component.tsx
```diff
@@ -34,6 +40,12 @@ const Plot = (props: PlotProps) => {
     YAxesSettings,
     XAxis,
     canvasRef,
+    gridVisible,
+    axesLabelsVisible,
+    xMinimum,
+    xMaximum,
```
louise-davies, line 46: I thought we had min/max settings in `XAxisSettings`/`YAxesSettings`? Makes fewer props to pass around, although I don't object to having them explicitly separate, but if the only option in the `(X/Y)AxisSettings` is `scale` then we should extract that too

mattjSTFC:
@sam-glendenning Let me have a mull over the data channel options and get back to you. One thing for sure let's try and steer away from multiple popups 👍

src/plotting/plot.component.tsx
sam-glendenning: You're right and I'd totally forgotten. I'm going to try things out by using the min and max fields in `AxisSettings`, it may well make things better

e2e/plotting.spec.ts
sam-glendenning: I took the dimensions out and it still worked fine! See commit https://github.com/ral-facilities/operationsgateway/pull/92/commits/2c3a0e4740123a8e2bdefc991c0e93f7e974bd75

e2e/plotting.spec.ts
sam-glendenning: Resolved by Louise in https://github.com/ral-facilities/operationsgateway/pull/92/commits/648c5ed1697c977545ced2d87b3bbaf8f8e4972a. This was caused by a config value not being set properly by Firefox, which has a workaround

src/plotting/plot.component.tsx
sam-glendenning: Tried it out but it proved too difficult to cohesively manage the state updates and callbacks. So I've extracted scale out to be its own type in https://github.com/ral-facilities/operationsgateway/pull/92/commits/3f43316c621f9d64db0329117b90118ee17a0f8d

e2e/plotting.spec.ts
sam-glendenning: Fixed in https://github.com/ral-facilities/operationsgateway/pull/92/commits/0aac476d20d999b153dfd3cb66ea6532a4f92dfc

sam-glendenning:
I'm assuming this will pass fine but Github is apparently experiencing "dragged performance" on Actions, according to Github Status. I'll come back to this tomorrow

mattjSTFC:
@sam-glendenning @louise-davies So rather than having all the options alongside a channel name we _could_ replace with a vertical dot icon button:

image

https://balsamiq.cloud/sq9hezy/p37ch0t/r45FA

Then surface a menu on the screen with the available options to style the plot (line style, colour and x or y) with the remove option set last:

image

https://balsamiq.cloud/sq9hezy/p37ch0t/r9DF1

I wasn't sure whether to add the visibility option to the menu (so all options are together) or, as per the wires, keep it near the channel name. The ability to quickly show / hide a plot line feels like it could be a **quick** action so having it within the menu may be annoying. What do you think?

**Pros**

- Frees up more space for the channel name
- Allows more channel options to be added without any concerns for limited space

**Cons**

- It may not be immediately obvious that there are more options by choosing the 3 dots icon. However if this pattern, of using the 3 dots to provide more option is applied elsewhere, then it should become clear to users. We also have a requirement for a Help section so ensuring this is as useful as possible will aid this way of working.
- What if only one option is surfaced as it makes the 3 dot pattern to show more redundant - not sure this will happen though as I imagine you will always have colour, line style and remove.

sam-glendenning:
Thanks for looking into this, @mattjSTFC. I'd like to add my input:
- I personally think that the user may want the "close" button as the quick action available to them rather than the visibility option. I feel it more likely the user will want to remove from the plot than temporarily hide data, hence my thoughts on that
- At the same time, I left out another useful bit of information. The user can also "hide" a channel from the plot by clicking its name in the legend above the plot (this is a feature of the plotting library we use, not something we built ourselves). However, unlike our own visibility option we programmed, this has the effect of rescaling the plot when done, whereas ours keeps the original plot scales. I'll try to make a video of this to show what I mean

mattjSTFC:
Thanks @sam-glendenning Happy to update the wires (with the close option upfront), especially as you are closer to it than me. Let me do that know 👍

louise-davies:
also, as a minor correction the axis is to choose between a left and right Y axis, not between the X and Y axis - but this shouldn't affect the design concept in any way. Also, we can potentially remove the line option style if a user is using a scatter graph (as obviously line style means nothing in that case) - so the "permanent" options are colour & axis, although there's potential for more configuration (e.g. point size)

mattjSTFC:
How about this:

image

image

sam-glendenning:
I'm personally a fan of it, although Louise should have the final input about the Y-axis radio buttons since she's working on it right now and it may not always be a binary choice (e.g. you can plot more than 3 channels but how this works in a plot with two y-axes I'm not entirely sure)

mattjSTFC:
Ok, I'll leave as is for now until I hear otherwise 👍

louise-davies:
Matt's design works well - re: Sam around the Y axis - you can have multiple channels on the same axis - the user just needs to be able to select which side. A radio button works fine for this

sam-glendenning:
Thanks for the clarification!

sam-glendenning:
Finally ready for another review again. Apologies, a lot has changed but it is mostly _moving_ existing code into better places, plus a few extra snapshots where required. I've added the new popup options menu as Matt suggested and refactored all of the plot settings code to their own separate components to improve the maintenance of the code going forward. Also a lot of the commits were me fiddling around with playwright trying to get it working.

src/plotting/plotSettings/moreOptionsToggle.component.tsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import { Box, IconButton } from '@mui/material';
+import { MoreVert } from '@mui/icons-material';
+import MoreOptions from './moreOptions.component';
+import type { MoreOptionsProps } from './moreOptions.component';
+
+// from react-colorful PopoverPicker demo :https://codesandbox.io/s/opmco?file=/src/PopoverPicker.js
+
+// Improved version of https://usehooks.com/useOnClickOutside/
+const useClickOutside = (
```
louise-davies: Isn't this the same code as in `colourPicker`? If it is, we should extract it to its own file to reduce code repetition

louise-davies (commented):
Functionality works well and this refactor will make it easier to modify parts of the plot settings in the future as they're split into smaller parts. Just the comment about the repeated code but otherwise it looks good :)

src/plotting/plotSettings/moreOptionsToggle.component.tsx
sam-glendenning: Fixed in https://github.com/ral-facilities/operationsgateway/pull/92/commits/3ba3a3c75f674481ecf0e9a8976807a326113d22

louise-davies (approved):

PR #100, opened 2022-10-11
DSEGOG-53 right y axis

louise-davies:
## Description
Allows users to switch channels between the left and right Y axis. The right Y axis is only visible if at least one channel is assigned to it.

Here's an screenshot showing the second Y axis in action, as well as showing the menu control

image

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes DSEGOG-53


sam-glendenning (changes requested):
Code looks good so far but see the below:

sam-glendenning:
Oh whoops, that didn't work. Let me try again

src/plotting/plot.component.tsx
```diff
@@ -121,7 +121,9 @@ const Plot = (props: PlotProps) => {
         },
         y2: {
           type: YAxesScale,
-          display: false,
+          display: selectedPlotChannels.some(
+            (channel) => channel.options.yAxis === 'right'
+          ),
```
sam-glendenning, line 126: We have this check for the right y-axis but could we also add the equivalent for the left one? If a user wants to, they could do this:

image

which looks a little confusing

sam-glendenning (changes requested):
Here we go:

src/plotting/plotSettings/moreOptions/moreOptionsBox.component.tsx
```diff
@@ -143,6 +169,53 @@ const MoreOptionsBox = (props: MoreOptionsProps) => {
...
+                componentsProps={{ typography: { sx: { fontSize: 12 } } }}
+                sx={{ margin: 0 }}
+              />
+              <FormControlLabel
+                value="right"
+                control={<Radio size="small" sx={{ padding: '2px' }} />}
+                label="Right"
+                componentsProps={{ typography: { sx: { fontSize: 12 } } }}
+                sx={{ margin: 0 }}
+              />
```
sam-glendenning, line 215: Could we add aria labels for these?

src/plotting/plot.component.tsx
louise-davies: We should check this with CLF - there's a specific bit in the requirements to hide the right Y axis if nothing is assigned to it but nothing for the left one

src/plotting/plotSettings/moreOptions/moreOptionsBox.component.tsx
louise-davies: These should all be accessible due to using labelled-by and the fact that radios are labelled by text

louise-davies:
@sam-glendenning should have probably only tagged you for review after I had put this out of draft - I haven't written any tests or anything 😆

src/plotting/plot.component.tsx
sam-glendenning: If we don't hide the left one, some values have to be assigned to it, which Chart.js does. And it potentially results in some confusion as it looks like one channel maps to two y-values. That was my reasoning

louise-davies: True, should probably just ensure that this case never occurs

src/plotting/plotSettings/moreOptions/moreOptionsBox.component.tsx
sam-glendenning: If you're happy, I'm happy. Didn't fancy busting out the screen reader for it ;)

src/plotting/plot.component.tsx
sam-glendenning: Either that or give the user the freedom to plot on the right y-axis while hiding the left one (I know this isn't usually done in practice but in case they wanted to)

src/plotting/plotSettings/moreOptions/moreOptionsBox.component.tsx
louise-davies: Inspecting accessibility properties on FF shows that all the roles and labels look fine & associated correctly

image

src/plotting/plotSettings/moreOptions/moreOptionsBox.component.test.tsx
```diff
@@ -122,4 +128,27 @@ describe('MoreOptionsBox', () => {
...
+        lineStyle: 'solid',
+        yAxis: i === 1 ? 'right' : 'left',
+      },
+    }));
+
+    createView();
+
+    await user.click(
+      screen.getByLabelText(`Pick ${props.channel.name} colour`)
+    );
```
sam-glendenning: Old copy-pasted code?

louise-davies: Oops - good spot! Yeah I just copied the previous test as the base of this one 😅

sam-glendenning (approved):

PR #101, opened 2022-10-14
DSEGOG-104 filtering

louise-davies:
## Description
Add the filter dialogue that lets the user contruct a filter and then passes that filter back to the backend via queries.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-104


sam-glendenning:
Have you had any inconsistencies with using the down arrow key to select the first option in the dropdown?

https://user-images.githubusercontent.com/71134574/196902965-6e931cd5-e010-436f-8736-aa0aadeaca13.mp4

louise-davies:
@sam-glendenning odd - the first item is automatically highlighted using the `autoHighlight` feature as this means the user can easily press enter to select the first option without having to press down. I'd expect pressing down should by default therefore select the second option as the first is pre-highlighted (this behaviour is the same as eCAT i.e. first option is auto-highlighted & pressing down selects the second option). The only way therefore that the down arrow would select the first option is if you somehow deselect it maybe? I can't recreate the behaviour on my side even though I followed the same steps (i.e. type s, press down, backspace, type s again, press up, press down, backspace, type s, press down -> this selects the second option)

sam-glendenning:
Ahh I see what the behaviour is supposed to be now - so the problem on my end is that for some reason the first option is not auto-highlighted. Therefore, I assumed it was skipping the first option. I'm not sure why it's not highlighting for me. I'll try another browser

louise-davies:
@sam-glendenning it is autohighlighting the first option most of the time - you can see in your video the top option is always "grey" i.e. selected. Pressing enter after typing "s" should select "Shot Number" without having to press down

sam-glendenning:
Having tried Edge, I do see it now. But there's quite a distinct difference between the original autohighlighting and then the highlighting of options as you move down the list with the arrow keys.

Autohighlighting (so light the screenshot can't even pick it up although I can see it more clearly on my screen):
image

Selected with arrow keys:
image

So for me, the problem is that the autohighlighting is not dark and obvious enough

louise-davies:
Cool, I should be able to change it so they're indistinguishable I would think

sam-glendenning:
And one more unrelated thing - how do I filter by time? I can't even figure it out on eCat

louise-davies:
I don't actually know! We should probably bring this up as a question to CLF as we don't have any guidance from behaviour from eCAT

EDIT: I suppose you should never have to filter by time, as you normally "search" by time instead via the datepicker

src/api/channels.tsx
```diff
@@ -63,7 +63,9 @@ export const generateChannelMetadata = (
 
 // TODO change this when we have a proper channel info endpoint to query
 // This just fetches metadata from the records endpoint at the moment
-const fetchChannels = (apiUrl: string): Promise<FullChannelMetadata[]> => {
+export const fetchChannels = (
```
sam-glendenning: Had a quick look for references elsewhere and I don't think we need this exported, since we can just use `useChannels`

sam-glendenning:
I guess that makes sense. But yeah we should definitely ask CLF because, if that is the case, isn't it pointless having `Time` as a selectable filter in the first place?

louise-davies:
@mattjSTFC another question for CLF - in eCAT you could select "Time" as a "filterable column" by it doesn't actually work as it doesn't accept a timestamp/date string. Thus, rather than handle timestamps/date strings it would make sense to remove Time as a filterable field, as the user can "filter"/search by time via the "Search" Epic/the datepicker at the top. However, as this is rechnically a change from eCAT and not in the requirements - we should double check with CLF that they are ahppy with us removing Time as a filterable column - as the alternative (correctly handling dates) is not impossible - just harder and  a bit redundant

src/filtering/filterParser.ts
```diff
@@ -0,0 +1,421 @@
...
+
+export class ParserError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'ParserError';
+  }
+}
+
+/** Class representing a stream of tokens, which keeps track of where in the array we've processed up to */
+class Input {
```
sam-glendenning: Could maybe rename this `TokenInput` to make it more distinguishable?

sam-glendenning:
Can I just confirm how you do a text filter? I typed in `Type` `=` `'GS'` and got no results (all the records have a `Type` value of `'GS'`

louise-davies:
Maybe a question for @MRichards99 or @kevinphippsstfc ? Just tried it myself and you're correct, but we're sending `conditions={"channels.Type.data":{"$eq":"GS"}}` which looks right to me - unless `eq` is only for numerical data...

sam-glendenning:
Should it be like DataGateway which uses 'ilike' and 'nilike'?

louise-davies:
But this is an exact string match so should work - it even matches case-wise. I'm not sure if there's a requirement for "like" - just "="

src/filtering/filterDialogue.component.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const { open, onClose } = props;
+  const dispatch = useAppDispatch();
+  const appliedFilters = useAppSelector(selectAppliedFilters);
+  const [filters, setFilters] = React.useState<Token[][]>(appliedFilters);
+  const [errors, setErrors] = React.useState<string[]>(['']);
+  const { data: channels } = useChannels({
+    select: (channels) => {
+      return (
+        channels
+          // TODO: I think it makes sense that we can only apply filters to scalar channels, but should check
```
sam-glendenning, line 52: Perhaps, although what if the user wants to do a null check for image and trace columns? Otherwise, a better function to use here would be our `useScalarChannels` function for plots so might as well use that instead

src/filtering/filterDialogue.component.tsx
```diff
@@ -0,0 +1,128 @@
...
+          // should also check we want friendly names here instead of system names
```
sam-glendenning, line 53: I agree, as these are displayed in the table if they're available

src/filtering/filterDialogue.component.tsx
```diff
@@ -0,0 +1,128 @@
...
+          <Grid item xs>
+            <Heading>Filter help</Heading>
+            <Body>
+              In the box, start typing data channel names, numbers, mathematical
+              symbols such as {'>'} and {'<='} and keywords such as AND and OR.
+              The Wizard will suggest suitable options and indicate using a grey
+              box when each item has been recognised.
+            </Body>
+            <Heading>Operators included</Heading>
+            <Body>{'>, <, =, !=, and, or, not'}</Body>
```
sam-glendenning: There's other ones too, like the parantheses and is null and is not null. Should we add these too?

src/filtering/filterDialogue.component.tsx
```diff
@@ -0,0 +1,128 @@
...
+      <DialogContent>
+        <Grid container columnSpacing={2}>
+          <Grid item xs pr={1}>
+            <Heading>Enter filter</Heading>
+            <Grid container item>
+              <FilterInput
+                channels={channels ?? []}
+                value={filters[0]}
+                setValue={handleChangeValue}
+                error={errors[0]}
```
sam-glendenning, line 93: Could you explain why only the first index of filters and errors is being passed here? In the FilterInput component, I console.logged the value prop to see how it was built up as the user added tokens. But what is supposed to go in the other indices of the filters object? For example, what do we intend to put in `filters[1]` and so on?

src/filtering/filterParser.ts
```diff
@@ -0,0 +1,421 @@
...
+    } else {
+      throw new ParserError('Missing operand');
+    }
+    this.op = input.consume(['compop', 'unaryop']) as Token;
+    if (this.op.type === 'compop') {
+      token = input.peek(0);
+      if (token !== null) {
+        if (token.type === 'channel') {
+          this.param2 = convertChannel(token.value);
+          input.consume();
```
sam-glendenning, line 246: Am I wrong or does this imply that there's a channel name on the right hand side of the expression? Can you do that? Can you even say something like `GEM_SHOT_NUM_VALUE` `=` `Shot Number`?

sam-glendenning (commented):
My thoughts so far (might leave the rest for tomorrow, as this is pretty heavy going)

Oh by the way, I have some leftover `Filter` types in `app.types.tsx` that can get cleared. I made them months ago when I copy-pasting a lot of the types from DataGateway in the anticipation that they would get used someday but as you've done the filters a different way, there's no need for them to be there. So they can get removed :)

src/api/channels.tsx
louise-davies: Ah yeah, I think I tried mocking it in a test but then just mocked axios instead - will revert

src/filtering/filterParser.ts
louise-davies: Yup, these are the names taken from eCat so there's not much logic behind them and I'm happy to rename any of them

src/filtering/filterDialogue.component.tsx
louise-davies: Multiple filter inputs is a separate issue - I just paid the groundwork for that here by making them arrays to begin with but just using the first index

src/filtering/filterDialogue.component.tsx
louise-davies: I transform the data into `Token`s in the `select` though, which wouldn't work with `useScalarChannels`. And as you said - we might actually want to be able to filter non-scalar channels... I'll leave the TODO as it is for now

src/filtering/filterDialogue.component.tsx
louise-davies: Yeah, I will update the help text 👍

src/filtering/filterParser.ts
louise-davies: Yes, in eCat you can compare columns to each other - but it looks like this isn't possible with the API right now. Is this a necessary feature? If so then @MRichards99 and @kevinphippsstfc need to know about it, if we don't know then we need to ask @mattjSTFC, and if we don't need to compare channels to one another then I'll modify this code to disallow it

src/filtering/filterDialogue.component.tsx
sam-glendenning: Oh yeah, because `useScalarChannels` is already a call to `useChannels` with its own options. That's fine

src/filtering/filterDialogue.component.tsx
sam-glendenning: Ah I thought so, I checked eCat and see they have the same setup

src/api/records.tsx
```diff
@@ -98,15 +102,25 @@ const fetchRecordCountQuery = (
...
+    [
+      string,
+      {
+        filters: string[];
+      }
+    ]
+  >
 ): UseQueryResult<T, AxiosError> => {
-  useAppSelector(selectQueryParams);
+  const { filters } = useAppSelector(selectQueryParams);
```
sam-glendenning, line 117: Just double checking - do we want the filters used in this function? I'm in the middle of changing this in my PR anyway because this function is only being used in plotting at the moment but what this means is that filters on the table also impact plot data. This also means that if a user draws a plot, minimizes the window, applies some filters on the table and opens the plot back up again, they may well find it's changed since they last looked at it!

louise-davies: This is how it works in eCat and how I think they want things to work. i.e. they use the table to select data and that is what is used in the plots

src/views/tableButtons.component.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    await user.click(screen.getByRole('button', { name: 'Filters' }));
+
+    const dialogue = await screen.findByRole('dialog', { name: 'Filters' });
+    expect(dialogue).toBeVisible();
+
+    await user.click(within(dialogue).getByRole('button', { name: 'Close' }));
+
+    await waitForElementToBeRemoved(() =>
+      screen.queryByRole('dialog', { name: 'Filters' })
```
sam-glendenning, line 42: Should we change this to `expect(screen.queryByRole('dialog', { name: 'Filters' })).not.toBeInTheDocument();`? Otherwise, would this return true if it were still in the document?

src/api/records.tsx
sam-glendenning: That's fine, just making sure

src/views/tableButtons.component.test.tsx
louise-davies: `waitForElementToBeRemoved` will fail is it's never removed (can be tested by commenting out the previous line clicking the close button) - it will error with `Timed out in waitForElementToBeRemoved.`. It's similar to how you don't need to assert that something exists when using `getBy*` queries - as RTL will fail if they don't exist.

cypress/integration/table/filtering.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    cy.contains('Apply').should('not.be.disabled');
+    cy.contains('Apply').click();
+
+    cy.wait('@getRecords').should(({ request }) => {
+      console.log('req', request);
+      expect(request.url).to.contain('conditions=');
+      expect(request.url).to.contain(
+        'conditions=%7B%22metadata.timestamp%22%3A%7B%22%24ne%22%3Anull%7D%7D'
+      );
```
sam-glendenning: Is there some way we can improve this by parsing some JSON or something down into this format without us having to write it ourselves? Saves us figuring out the percentages and stuff

sam-glendenning:
I think that's all the comments I have to make, everything looks good otherwise!

cypress/integration/table/filtering.spec.ts
louise-davies: Yeah, I can do some URL escape conversion 👍

louise-davies:
@sam-glendenning I've now fixed the not behaviour to take into account that the mongoDB `$not` operator only works before an operator. Now, we propagate the `NOT` to the `Predicate`, which inserts the mongoDB `$not` operator in the appropriate place. This takes advantage of De Morgan's Laws i.e `NOT (A AND B) == NOT A OR NOT B` and `NOT (A OR B) == NOT A AND NOT B` so you can see there's also logic to swap the and/or's when appropriate. Do have a play about - it does get confusing quite quickly but in my tests it all seems to work!

src/filtering/filterParser.ts
```diff
@@ -357,10 +360,20 @@ class BooleanTerm {
...
+        s += '{"$and":[';
+      }
       this.factors.forEach((f) => (s += f + ','));
       s = s.slice(0, -1);
-      s += ']}';
+      if (this.not) {
+        s += ']}';
+      } else {
+        s += ']}';
+      }
```
sam-glendenning: Okay the same thing is happening here and since I'm struggling a little to figure out everything going on, are you sure this is correct? We don't want to accidentally have the opposite expression of what we're aiming for

sam-glendenning:
Pending a reply to the above, I think the changes to NOT are correct (I did need a review of De Morgan's Laws, as it's been some time). The logic in `filterParser` can indeed get very tricky thanks to nested logic operators flying around everywhere but I think you've implemented it well. Fortunately, the unit tests are very straightforward and seem to confirm that everything runs as expected

src/filtering/filterParser.ts
louise-davies: Ah, this was a hold over from when I was doing something different - I'll delete this. Previously, I was trying to use `nor` and `nand` - but there's no `nand` operator in mongoDB, instead you can create the same effect with a `nor` wrapping an `and` - hence why in the `not` case here I needed different "closing brackets". However, that logic was more confusing/conflicted with being able to `not` normal expressions, so I switched to this method but forgot to remove these lines

sam-glendenning:
Yep and one more thing, a reminder to remove the old Filter types from `app.types.tsx`

src/filtering/filterParser.ts
sam-glendenning: Following a discussion on Slack, this will be tracked in a Jira issue once we have clarification from CLF

src/app.types.tsx
```diff
@@ -113,7 +113,7 @@ export interface SortType {
 
 export interface QueryParams {
   sort: SortType;
-  filters?: FiltersType;
+  filters: string[];
```
sam-glendenning: Sorry I know I keep banging on and still haven't approved this but this change has actually illuminated another problem to me. At first I saw this and thought "how was this even working before?" but then realised that this interface isn't actually being utilised anywhere, not even in `fetchRecords()`. So what we could do here is either stick with that but remove this interface altogether or we could actually start making use of it and integrate its use into `records.tsx`

louise-davies: I contemplated removing it as we can infer the type from `selectQueryParams` - I'll just removing it for now as we're not using it. We can always add an interface back in if it's necessary.

sam-glendenning (approved):
Great addition, makes some solid improvements!

louise-davies:
I love a good Friday afternoon PR approval

PR #102, opened 2022-10-18
DSEGOG-174 saving plots to redux store

sam-glendenning:
## Description
This allows for the user to save the plot they have open in a window to the Redux store. This lasts while the main window is not closed or refreshed, since sessions have yet to be setup.

The Save As button has been removed as I didn't deem it useful. Instead, the Save button saves the plot under the title it currently has. If a plot is renamed and saved, a new saved plot is stored in Redux, with the old copy remaining under the old name.

When the user saves a plot and closes it, it can be reopened with its state restored from Redux.

## Testing instructions
Try saving a plot and reopening it with the Edit button. All state should be restored.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes DSEGOG-174


sam-glendenning:
Reminder for Sam to try and add the plot window position on screen to the plot saved state

sam-glendenning:
One thing I overlooked - I forgot to implement functionality to restore the window position as it was. Is this still a desired feature?

louise-davies:
@sam-glendenning This is a feature on the wishlist & is part of one of the sessions requirements so I think it's still desired yes

src/plotting/util.tsx
```diff
@@ -0,0 +1,13 @@
+import { SelectedPlotChannel } from '../app.types';
+
+// TODO test this
+export const deepCopySelectedPlotChannels = (
```
louise-davies, line 3: If you want to actually deep copy - you're best using `JSON.stringify` and `JSON.parse` - this method works fine as long as the options in `options` are primitives and not objects themselves (as otherwise those underlying objects are still the same by reference)

louise-davies (commented):
All the plot functionality still seems to work and saving a plot works and reopening that plot renders the same plot. I'm just not sure of the UX of "creating" a new plot automatically when the user saves a plot - especially in the case when it changes from `Untitled X` to a named plot.

I wonder if we do bring back the `Save As` button which does what the current Save button does, and have the `Save` button overwrite the plot config that was used to open the plot... Or have a dialogue pop up asking you whether you want to overwrite or create a new plot...

However, I'm happy for it to go in as is and we can change this later if needed, as I'm aware that Matt Jones is on leave so we can't ask him until next week

src/plotting/util.tsx
sam-glendenning: That was an alternative I considered. Happy to make the change

sam-glendenning:
I wasn't sure of this either. Any approach can be done, I'm just completely unsure what the user might want. I went with this as it was the safe option in case the user didn't want to lose an old plot, plus it acts as a method to "copy" plots in a sense? Anyway, shall we tag Matt and ask his opinion for when he does come back?

src/plotting/util.tsx
louise-davies: It's probably best to make it to prevent any future accident problems by the fact that this is only a single level deep copy. I had similar problems with the 2 y axis e2e test where I was accidentally modifying the underlying mock request data instead of deep copying because I just did object spread instead of JSON stringify&parse

sam-glendenning: Resolved in https://github.com/ral-facilities/operationsgateway/pull/102/commits/19fbd764a1ca526d887c958298bcb5adcc5454c7

louise-davies (dismissed):
Looks good. Just some questions about the save method, but we can change that in a later PR if needed. @mattjSTFC ping for when you get back but we want to know how the save plot behaviour is expected to work

sam-glendenning:
You will likely have to re-approve, I'm in the middle of trying to save the window position as well. I'll let you know

sam-glendenning:
The window size and position are now also saved when the save button is clicked. It works _fairly_ well, certainly on a single monitor, although on multiple monitors your mileage may vary. If you have any ideas on how to improve this, fire away although I think it will be challenging.

louise-davies:
@sam-glendenning I did say when this requirement was discussed that we could only do our best and that MMV depending on browsers/monitor setups, so this looks fine to me

sam-glendenning:
@louise-davies are you happy to approve?

louise-davies (approved):
As you say - the window position saving works pretty well if the monitor you have the OG main window open on is the same as the one you have the plots on. Otherwise it uses the wrong screen (in Chrome) or uses the wrong aspect ratios (FF) - but I think this is acceptable because a) there's literally not much else we can do re: controlling the size and position of the popup window and b) it's better than nothing, especially in the single window case.

Again, there's some alternative ideas on how saving plots should work re: overwriting existing plots vs creating new plots, but we can address those later

sam-glendenning:
Thanks for the extended testing on other browsers, I admit I had not done that. Ultimately, we'll see what end users think!

PR #109, opened 2022-10-25
DSEGOG-105 multiple filters & deletion

louise-davies:
## Description
Adds an "Add new filter" button which creates extra filter boxes, and a delete button next to each filter input to delete said filter input.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-105, partially closes DSEGOG-107

### DIFF
--- a/src/filtering/filterDialogue.component.test.tsx
+++ b/src/filtering/filterDialogue.component.test.tsx
@@ -118,4 +118,94 @@ describe('Filter dialogue component', () => {
 
     expect(screen.getByText('Apply')).toBeDisabled();
   });
+
+  it('adds new filter when Add new filter button is clicked and renders multiple filters', async () => {
+    const user = userEvent.setup();
+
+    createView();
+
+    await user.click(screen.getByText('Add new filter'));
+
+    expect(screen.getAllByRole('combobox', { name: 'Filter' })).toHaveLength(2);
+  });
+
+  it('deletes a filter when delete button is clicked', async () => {
+    const state = {
+      ...getInitialState(),
+      filter: {
+        ...getInitialState().filter,
+        appliedFilters: [
+          [
+            { type: 'channel', value: 'timestamp', label: 'timestamp' },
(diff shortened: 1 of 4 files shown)


src/filtering/filterDialogue.component.test.tsx
```diff
@@ -118,4 +118,94 @@ describe('Filter dialogue component', () => {
...
   });
+
+  it('adds new filter when Add new filter button is clicked and renders multiple filters', async () => {
+    const user = userEvent.setup();
+
+    createView();
+
+    await user.click(screen.getByText('Add new filter'));
+
+    expect(screen.getAllByRole('combobox', { name: 'Filter' })).toHaveLength(2);
```
sam-glendenning, line 129: Could we do a check for what happens when the Apply button is clicked here as well? i.e. check the store holds both filters

sam-glendenning (approved):

sam-glendenning (changes requested):
Wait go back, I meant to press request changes

src/filtering/filterDialogue.component.test.tsx
louise-davies: This test is just about that it adds an additional filter box though. If I just press `Apply` here, the store won't update as they're both blank and we remove "extra" blanks. I could modify the `delete a filter` test so that there's three initial filters and that after I delete one there's two instead? Otherwise this test would have to be much longer as I would have to type out the filters to get them to save...

louise-davies: I test that you can apply multiple filters in the e2e tests as well

sam-glendenning: Right, my meaning was to actually extend the test and add multiple, actual filters to the store. I do think it's a worthwhile test to have

louise-davies: Okay, I'll do  that then

sam-glendenning (approved):

PR #110, opened 2022-10-26
DSEGOG-106 adding column filtering indicator

sam-glendenning:
## Description
This adds a filter icon in the header of a column that is currently subjected to filtering. Strictly speaking, any `Token` of type `channel` in the `appliedFilters` variable is marked as subject to filtering, so if it appears in the record table, it will have a filter icon next to its header.

The wireframes also mention that it would be nice if this icon opened the filters dialogue when clicked. So I've added this as well, with the token(s) in question being highlighted briefly when the dialog box opens. To allow for this, I had to rework a lot of the functionality that allows for the filters dialog to open and lifted it into the `dataView` component.

Further changes include some updates to some existing tests to make them better.

## Testing instructions
Mention channels in a filter input and they should appear with a filter icon when the user views that column in the table. This icon should open the filter dialog when clicked, highlighting the channel in question for a short period of time.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-106


louise-davies (changes requested):
I noticed a few issues re: the table
1. Toggling word wrap on & off or reordering columns now resets all the column widths (doesn't happen on develop)
2. The hover on both the column options menu and the filter icon are not circular: see 
image (sorry for the faint screenshot)

sam-glendenning:
I have fixed the alignment of the buttons in the data header as best I can. The intention behind it is to fix the issue where the buttons are too spaced apart by default with no styling. There were two solutions I could go with to fix the highlighted issue:

Option 1 - hover shadows are centrally aligned again but the actual icon appears lower down than the menu icon:


Option 2 - icon is horizontally aligned with the menu icon at the cost of the hover shadow being partially cut off at the top (problem being that the hover shadow of the filter icon and the menu icon are not the same size):


I decided to go with option 2, as the shadow alignment issue is less noticeable to the end user

sam-glendenning:
I've fixed the issue of the column widths resetting. It was caused by the `availableColumns` resetting whenever the user toggled word wrap, added/removed a column, etc. Why exactly, I'm not sure. Possibly because the filters being passed to `useAvailableColumns` were causing a refetch of data every time? Anyway, I've moved everything back to where it was in `channels.tsx` and now parse the filtered channel names in `recordTable.component.tsx`.

A column no longer has a `filtered` flag in its column instance anymore as a consequence of this. So there's more of a separation between the filters and the columns now but this was how I got things to work. I could potentially add a `filtered` flag to the `ColumnState` interface but this would require modifications of the table slice and multiple `dispatch` calls to make the table and filters state cohesive and I just didn't see the point. Let me know your thoughts.

louise-davies:
@sam-glendenning I've managed to make the focus ripple circular and keep things nicely aligned - you just need to wrap the button in a div. I think this is because the button isn't forced to be square but will default to square in a div (or something!). It was either that or set explicit sizes for the button which imo is worse for maintainability.

image

If you're happy I can commit the changes :)

sam-glendenning:
Oh nice! That would be gem, thank you

louise-davies (approved):
Works well, previous issues have been fixed & I really like the highlight when you click the indicator :)

PR #116, opened 2022-11-02
DSEGOG-69 DSEGOG-89 attempt to fix the date-time pickers

sam-glendenning:
## Description
This attempts to improve the existing date-time pickers functionality. 

- I've harmonised the way we capture and update date parameters for the search query with the way DataGateway does it. This was originally an attempt to fix the behaviour of the DateTimePickers being partially updated in the picker menu and not updating the query but this didn't solve the actual problem
- Fixing a bug where pressing the Enter key reloaded the page. This was because of the surrounding `form` tag, which I've replaced with a `Grid` to get the same effect.
- Adding CLEAR, CANCEL and OK buttons. The picker NEARLY works perfectly, but I can't achieve the same functionality with the OK button as in DataGateway. It seems the entirety of the field needs to be completed in the picker before the value is accepted. This is not true if the user inputs the value manually. 

So essentially - the OK button is useless in a DateTimePicker! Because, regardless, it doesn't make the picker accept what is currently filled in the input field. These are issues that [others have noticed](https://github.com/mui/mui-x/issues/4599) and developers [deliberately want this behaviour to continue for various reasons](https://github.com/mui/mui-x/issues/6161).

As a result, I've had to stick with the existing combined usage of the `onChange` and `onAccept` functions to achieve a decent amount of the functionality we want between the picker and the text input. So I'm eager to get some input from Louise on this one. We can either:

- stick with the existing functionality and I remove the pointless OK button
- we use the functionality provided by the developers where the `onChange` method is used to update the search parameters every time it calls (this means the table will update for each click of the user selecting date, hour, minute and second)
- some other option I haven't thought of?

Bottom line - it seems DatePickers behave quite differently from DateTimePickers.

## Testing instructions
Check that pressing Enter no longer reloads the page. Check the CLEAR and CANCEL buttons work as intended. Suggest anything else that might be improved.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes DSEGOG-69, DSEGOG-89


sam-glendenning:
I have refactored the code quite a bit so it's important to make sure the `onChange` method is still only being called when we want it to be. We want it to be called in the following scenarios:
- fromDate given a valid date
- fromDate being cleared
- toDate given a valid date
- toDate being cleared

So test, test and test again. Try and break it!

Final note: this has a low code coverage level due to the introduction of a couple more lines of code to make the handling of date inputs between the text field and the picker more stringent. As I still haven't got a good way of testing the picker itself, these lines remain uncovered.

louise-davies:
@sam-glendenning I believe the problem with the OK button not working is the `if (datesEqual(date, prevDate)) return` check in `updateFilter`. Because you're passing in the "temporary" date that gets changed in intermediary steps, this line prevents the OK button from actually submitting a date. Changing calls to `updateFilter` to instead use the previous date from the props (i.e. the last date the user submitted) - then the OK button seems to work fine for me? Instead there was some weird requerying happening sometimes on cancellation, but that's less of an issue I think right?

OG-datetime-pickers

In the code, I have defined
```
const initialFromDate = props.value?.fromDate
    ? new Date(props.value.fromDate)
    : null;
const initialToDate = props.value?.toDate
    ? new Date(props.value.toDate)
    : null;
```

and these are passed in to the `updateFilter` in the `prevDate` slot instead of `fromDate` and `toDate` respectively.

sam-glendenning:
Ahhh interesting. I did think something was a bit fishy but couldn't quite wrap my head around it. If you want, you're welcome to commit that change and I can test it out on Monday as well. I would much prefer asking the user to "confirm" their choice by having to press OK in the picker. 

Then again, this may change down the line because the wireframes show a SEARCH button at the end of the search fields which we could program to actually launch the search query and prevent any input into the other search fields launching the query themselves whenever they change. In this instance, if we change the date fields or the shot ID or whatever, the search query doesn't launch at all until the user presses the SEARCH button. Still working on this so if you have any thoughts on that by all means share them.

sam-glendenning:
@louise-davies ready for another review

src/views/dateTimeInput.component.tsx
```diff
@@ -30,49 +30,51 @@ export function updateFilter({
...
 
   const invalidDateRange = fromDate && toDate && isBefore(toDate, fromDate);
 
+  const initialFromDate = props.value?.fromDate
+    ? new Date(props.value.fromDate)
+    : null;
+  const initialToDate = props.value?.toDate
+    ? new Date(props.value.toDate)
+    : null;
+
```
louise-davies: Can we move these earlier and use these to initialise `fromDate` and `toDate`? Just reduces some repetition

louise-davies (approved):
I've tested this and the new changes make it so the `Cancel`, `Clear`  and `OK` buttons all work as expected :)

PR #120, opened 2022-11-08
Updating loader-utils to v3.2.0

sam-glendenning:
## Description
We recently had a couple of PRs on [OperationsGateway](https://github.com/ral-facilities/operationsgateway/pull/119) and [DataGateway](https://github.com/ral-facilities/datagateway/pull/1455) created by Dependabot to resolve critical vulnerabilities recently discovered in [v2](https://github.com/webpack/loader-utils/issues/212) and [v1](https://github.com/webpack/loader-utils/issues/218) of loader-utils respectively. I have since merged both PRs but neither project is using the latest version of loader-utils.

This PR aims to address that by upgrading to v3. It may be I'm unaware of why we're using an older, deprecated version (perhaps it's still a transitive dependency for many of our dependencies) so if that's the case, we can close this PR. Otherwise, we can upgrade and consider doing the same with DataGateway.

## Testing instructions
Check latest version is being used.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Hotfix


sam-glendenning:
My conclusion is that this is fundamentally incompatible with some essential libraries we use, therefore I'm happy to stick with v2 which, while deprecated, is still heavily used in the community. I'll close this PR.

PR #127, opened 2022-11-14
DSEGOG-95 search by shot number

sam-glendenning:
## Description
Allows the user to search by shot number through a popup window containing min and max values, as per the wireframes. I've refactored the functionality behind searching as a whole so that the existing date-time search functionality works in tandem with it. As a result, there are a couple of breaking changes:

- search is launched when the user presses the big shiny SEARCH button. This replaces previous behaviour in the date-time pickers which updated the search parameters whenever the user updated their values. Whatever values are currently in the search fields will be used in the new search query and the table won't be updated until the search button is pressed again with new search entries
- a need to tie the search parameters in with the existing filters work mean all the search and filter conditions passed to the backend are wrapped in one big `$and` statement. As far as I can tell, this works as expected but please test this yourself
- The wireframes seem to indicate that min and max fields for searching by a shot number range are always present but I may have this wrong. Therefore, as it stands, a min and max parameter for shot number of 0 and 99999999 respectively are included with every query (unless the user changes them). This has complicated queries and unit tests. It may be that we decide to make these optional, in which case I'll have to change the design a bit

## Testing instructions
Test searching _and_ filtering on various fields and check the data being displayed matches what you expect.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes DSEGOG-95


sam-glendenning:
Following a conversation, we have decided to make the min and max fields optional. They can be greyed-out "default" values in the input boxes but not selected unless the user edits them

sam-glendenning:
@louise-davies ready for a further review with the above changes implemented (optional shotnumber range)

src/search/components/dateTime.component.tsx
```diff
@@ -20,100 +19,95 @@ export interface UpdateFilterParams {
...
-            // onChange handles user-input date-time changes
             onChange={(date) => {
               setFromDate(date as Date);
-              if (datesEqual(date, fromDate)) return;
-              if ((date == null || isValid(date as Date)) && !popupOpen) {
+              if (!popupOpen) {
                 updateFilter({
                   date: date as Date,
-                  prevDate: initialFromDate,
+                  prevDate: fromDate,
```
louise-davies: This reintroduces the bug we had previously, where "half" selecting a date doesn't work. We need to use `receivedFromDate`/`receivedToDate` in `prevDate`. See gif

OG-timestamp-search-bug

Also, as an aside, the box for the from date/to date is overly large on my monitor, with a lot of deadspace on the right

louise-davies (changes requested):
The shot number functionality works well - just need to fix the issue with the date/time pickers

src/search/components/dateTime.component.tsx
sam-glendenning: Thanks for pointing that out. I'm annoyed I let that back in there. And the deadspace issue - a constant reminder I need to work on my CSS skills

louise-davies (approved):

PR #138, opened 2022-11-25
DSEGOG-98 DSEGOG-99 search by predefined and custom relative timeframes

sam-glendenning:
## Description
This adds to the search bar by allowing the user to search by predefined and custom relative timeframes. It includes buttons to search by last 10 mins, last 24 hours and last 7 days. Below this, it includes an input box to enter a user-specified number of minutes, hours or days.

When these buttons are clicked, the date-time values in the date-time pickers are updated accordingly. The timeframe itself is not added as part of the query: it is solely used to change these date-time values.

## Testing instructions
I've done my best to ensure that date-time values can be updated in the date-time pickers themselves and the timeframe component. This required multiple variables in different component states to keep track of things but we need to make sure this is all working before merging this. Please test this as much as possible. I've added comments where I can to help explain what's happening.

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes DSEGOG-98 DSEGOG-99


src/search/components/dateTime.component.tsx
```diff
@@ -14,27 +14,22 @@ export const datesEqual = (date1: Date | null, date2: Date | null): boolean => {
...
   date,
   prevDate,
   otherDate,
   fromDateOrToDateChanged,
   changeDate,
-}: UpdateFilterParams): void {
-  if (
-    date &&
-    isValid(date) &&
-    !datesEqual(date, otherDate) &&
```
louise-davies, line 35: We've removed this check here - from what I can tell this stops us from having identical dates in the from and to picker. Is there a reason this was removed? I suppose this case is caught by the `isBefore` checks below

src/search/searchBar.component.tsx
```diff
@@ -12,35 +14,101 @@ import {
...
+    switch (timeframe.timescale) {
+      case 'minutes':
+        from.setMinutes(to.getMinutes() - timeframe.value);
+        break;
+      case 'hours':
+        from.setHours(to.getHours() - timeframe.value);
+        break;
+      case 'days':
+        from.setDate(to.getDate() - timeframe.value);
+        break;
```
louise-davies: We do have `date-fns` available - `sub` may be useful but since it's relatively simple maths I think we're fine with native functions - if we had to be able to do time frames like "1 day and 1 hour ago" then I think using `sub` would be best, but I don't think we need that.

src/search/searchBar.component.tsx
```diff
@@ -12,35 +14,101 @@ import {
...
+  // This is set in the timeframe search box by the user
+  // If the user edits any of the date fields otherwise, this is set to null
+  const [timeframeRange, setTimeframeRange] =
+    React.useState<TimeframeRange | null>(null);
+
+  const calculateTimeframeDateRange = (
+    timeframe: TimeframeRange
+  ): { from: Date; to: Date } => {
+    const to = new Date();
+    const from = new Date();
```
louise-davies: I think this is the reason perhaps why you're sometimes off by a second in tests - theoretically between these two lines the seconds could tick over. If we change the definition of `from` to be `const from = new Date(to);`, this ensures that a) what we consider "now" is the same in both `to` and `from` and b) `from` is not a reference to `to` and is instead a new date object with the same time info as `to`

src/search/searchBar.component.tsx
sam-glendenning: I hadn't thought about using that, I think it's a good shout! Much more reliable

src/search/searchBar.component.tsx
sam-glendenning: That might help, I knew the reason but didn't seem to apply the resolution globally, even though I have used that method in other parts! I'll give it a go

src/search/searchBar.component.tsx
louise-davies: It also means we don't need the case statement - we can feed the units directly into `sub`

src/search/components/dateTime.component.tsx
sam-glendenning: Do you mean the `datesEqual` function? I think it's still there, no? Line 32

louise-davies: We don't check `!datesEqual(date,otherDate)` anymore, we're still checking `!datesEqual(date,prevDate)`

src/search/components/timeframe.component.tsx
```diff
@@ -1,17 +1,142 @@
...
+          </Button>
+        </Grid>
+        <Grid item xs={4}>
+          <Button
+            variant="outlined"
+            sx={{ height: '100%' }}
+            onClick={() => changeTimeframe({ value: 7, timescale: 'days' })}
+          >
+            Last 7 days
+          </Button>
```
louise-davies: Can we set `size="small"` on these buttons too? I don't think they need to be the bigger size, and it just adds a little consistency to all the button's padding etc in the popup

src/search/components/dateTime.component.tsx
louise-davies: Yeah I just checked, if `from` and `to` are the same then it's now a valid date range - this could be fine though as just checking we use `gte` and `lte` in the query to the API, so now we accept being able to search for a specific second.

sam-glendenning: Ah sorry, I realised my mistake the second I wrote that. I removed that because I was trying to allow the user to search by two equal dates so they can see shots in a one-second time interval. But then I cancel that out with the `isBefore` below so clearly I didn't think straight there. But none of this seems to work anyway because I _can_ search by equal dates at the moment?

While testing this, I've also noticed another bug which I've found a fix for, which was that copy pasting dates into the fields didn't work. So with all of that said, I'm going to convert this to draft and work on it further

src/search/components/timeframe.component.tsx
sam-glendenning: I was debating that too, I'll go for it

sam-glendenning: Addressed in https://github.com/ral-facilities/operationsgateway/pull/138/commits/a4a01470deab1f77015f18d40ad31b027b422d1d

src/search/searchBar.component.tsx
sam-glendenning: Addressed in https://github.com/ral-facilities/operationsgateway/pull/138/commits/793b2ad3b408be6e047f6895dcb6bb415a4023da

src/search/components/dateTime.component.tsx
sam-glendenning: Addressed in https://github.com/ral-facilities/operationsgateway/pull/138/commits/66b9877748bb3bd79cd73955c89e8cabf373b85b

src/search/searchBar.component.tsx
sam-glendenning: @louise-davies okay having gone back over things, I now remember the exact problems. The issue stems from the fact that when I calculate the `expectedDate` in tests and when the test clicks the relative timeframe button (and calculates the current date when THAT happens), those two date objects MAY differ by one second due to happening in two different seconds.

I am unsure how to address this in unit tests.

I attempted to address this in e2e tests through the use of `cy.clock()` and `clock.setSystemTime()` but these are only available in Cypress ^v10.7

So I have not resolved anything in this regard

louise-davies: In unit tests you could mock the `Date` constructor to always give the same date if you provide it with zero arguments I suppose

`cy.clock` can be used without `setSystemTime` to mock `Date` functions - see https://github.com/ral-facilities/datagateway/blob/e893e0e49ed5b5b32442b95e29b273e16b3ad764/packages/datagateway-download/cypress/integration/downloadStatus.spec.ts#L39

sam-glendenning: I will give this a go

sam-glendenning: I have managed to get this working with unit tests but mocking the time across e2e tests I haven't. I believe it's because you can mock what Cypress believes the time to be but you still have the problem with the built-and-running application itself still using the current system time

src/search/components/dateTime.component.tsx
```diff
@@ -14,30 +14,30 @@ export const datesEqual = (date1: Date | null, date2: Date | null): boolean => {
...
   otherDate,
   fromDateOrToDateChanged,
   changeDate,
-}: UpdateFilterParams): void {
+}: VerifyAndUpdateDateParams): void {
   if (
     date &&
     isValid(date) &&
-    !datesEqual(date, otherDate) &&
+    // !datesEqual(date, otherDate) &&
```
louise-davies: Remove this commented line if it's not useful anymore

sam-glendenning: Addressed in https://github.com/ral-facilities/operationsgateway/pull/138/commits/429360b8c1be31137242057ba69c5e9b465695a3

louise-davies (approved):

PR #142, opened 2022-11-28
DSEGOG-202 updating tooltip mode to 'point'

sam-glendenning:
## Description
This moves the tooltip next to the point the mouse is hovered over. This changes the tooltip position from this:
image

to this:
image

## Testing instructions
Compare the above screenshots or test yourself!

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-202


e2e/plotting.spec.ts
```diff
@@ -223,7 +223,7 @@ test('user can zoom and pan the graph', async ({
   }
 });
 
-test('plots multiple channels on the y axis and displays correct tooltips', async ({
+test('plots multiple channels on the y axis', async ({
```
louise-davies, line 226: So are we happy with not having a test that the tooltips are working? It's fine not to - as that's functionality from Chart.js that we're not modifying really, and it would now be much harder to test here as we'd need an exact pixel position for the mouse to hover over.

sam-glendenning: Yeah it's annoying. As I said in my commit message:

"Due to the more precise way tooltips are now rendered, it is practically impossible to get them to appear in playwright tests anymore. This is because the user needs to precisely line their mouse up over a point to get a tooltip to appear and despite trying to make this happen in tests I couldn't manage it. I was also working blind because the canvas element itself does not appear in playwright trace mode but does in snapshots"

Unless you have a magic wand to make this work, I can't see a workaround

louise-davies (approved):

PR #143, opened 2022-11-28
DSEGOG-103 make search box collapsible

louise-davies:
## Description
Add a button to expand/collapse the search box

There's no wireframe/mockup showing the expand/collapse functionality, so I've just put it with the other buttons (e.g. `Filters` etc.) but this can easily be moved

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-103

### DIFF
--- a/cypress/integration/table/search.spec.ts
+++ b/cypress/integration/table/search.spec.ts
@@ -15,6 +15,18 @@ describe('Search', () => {
     cy.visit('/').wait(['@getRecords', '@getRecords', '@getRecordCount']);
   });
 
+  it('can be hidden and shown', () => {
+    cy.contains(/^Search$/).should('be.visible');
+
+    cy.contains('Hide search').click();
+
+    cy.contains(/^Search$/).should('not.be.visible');
+
+    cy.contains('Show search').click();
+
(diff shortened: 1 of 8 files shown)


sam-glendenning (approved):
Works well! I wonder, what's your thoughts on having a little chip to the right of the words 'Hide/Show Search' that contains a number equalling the number of active search parameters? For example, if we have a shot number range selected, the chip contains 1. If we have a shot number and date range selected, the chip contains 2. If none are selected, there is no chip.

The purpose is purely to highlight that there are search parameters currently active when the user has the search collapsed and may have forgotten these are active

louise-davies:
@sam-glendenning sounds like a neat idea but I'm not sure how it would work with the idea of correlated fields - e.g. searching by experiment id will automatically adjust the date range to the date range of the experiment - does that mean the user has filtered by date range or not? I'd stick with what we have for now, and either @mattjSTFC or users can suggest improvements

sam-glendenning:
Yes I had similar concerns with the timeframe adjustment. I'll approve what's here since it's good as is

sam-glendenning:
Actually, I've already approved. Merge when ready!

PR #144, opened 2022-11-30
DSEGOG-101 specify max shots to display

sam-glendenning:
## Description
This adds radio buttons to allow the user to specify the maximum number of shots
- to display in the plot
- to paginate through in the record table

The current setting for max shots is stored as part of the search parameters, meaning things only update when the user clicks Search.

## Testing instructions
There's only so much testing of this that can be done considering the options for Max Shots are 50, 1000 and Unlimited while the test data we have is limited to just over 100 shots. However, I've mocked things where I can in tests. At the moment, when you click on 1000 or unlimited, the correct logic should kick in so that the table and plots display all shots.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-101


src/app.types.tsx
```diff
@@ -124,6 +124,7 @@ export interface ShotnumRange {
 export interface SearchParams {
   dateRange: DateRange;
   shotnumRange: ShotnumRange;
+  maxShots: number | 'Unlimited';
```
louise-davies: I wonder if some of the logic here would be simplified if "Unlimited" `maxShots` is represented by `Infinity`. It would mean no special handling for the table - as `totalDataCount < maxShots ? totalDataCount : maxShots` works if `maxShots` is `Infinity`. Also means we don't have to deal with both numbers and strings in parsing the values in `maxShots.component.tsx` - the only change is instead of using `parseInt(value)` we'd have to use `Number(value)`. Thoughts? Feel free to ignore if you try it and the logic is more complex

louise-davies (commented):
Tested it and it's working fine with both the table and plots. Just my comment on using `Infinity` instead of a string and I'm happy for it to go in :)

src/app.types.tsx
sam-glendenning: Use of `Infinity` had not crossed my mind but I like it and it does simplify things. I'll make the change

sam-glendenning:
There are definitely conflicts between this and https://github.com/ral-facilities/operationsgateway/pull/138 so approve one before the other

louise-davies (approved):

PR #149, opened 2022-12-06
DSEGOG-100 tooltip warning message over search button if record count > limit

sam-glendenning:
## Description
This adds a warning message in the form of a tooltip over both the Search button and the filter dialogue Apply button if we detect the new search will return more records than the limit set in the settings file.

This is done through:
- creating a new `recordLimitWarning` setting in `operationsgateway-settings.json` to define the maximum number of records before this warning is displayed (set to -1 for unlimited)
- using an `incomingRecordCount` query to get the `count` of records in the new search before actually fetching those records
- if the count is greater than the limit set, a tooltip is displayed over the Search or Apply button (whichever is handling the search at that moment)
- the search is then only "confirmed" and initiated when the user clicks the button for a second time

There's had to be a significant rework to get this all to occur. ~~I've not yet found a way to transfer the incoming record count into the actual record count itself, meaning two requests for the count occur for each declared search. This isn't ideal.~~

I've also exposed a problem where there are two methods of searching (through the search components and the filter dialogue) but each only takes into account its own localised parameter changes. What this means is, if I change the shot number range in search, then add a filter and hit Apply, only the new filter parameters have been applied to the search while the shot number range is present in the shot number search component. So we'll need to find a solution to this down the line.

~~I would have liked to have found a way to not display the tooltip if it had already been displayed for a particular set of search parameters but was unable to find a reliable way of doing this.~~

Currently, the `recordLimitWarning` setting is mandatory and given a positive integer for an actual limit setting and -1 to denote unlimited but I can change this to make the setting optional instead.

I attempted some e2e tests for the tooltips displaying and the count request being made but I think the two count requests occurring has got in the way of this and I couldn't get very far with it. Fortunately, unit tests cover tooltip tests quite nicely.

So overall, this could use some improvement.

## Testing instructions
Tests should hopefully pass, it's just whether or not we want this specific implementation.

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes DSEGOG-100


sam-glendenning:
The search e2e tests need the recordLimitWarning to be set to -1 (hence the stubbed settings response) and an additional wait for an extra count request being made (I'm not sure if this is the incomingRecordCount query or a bug). But for some reason, the filtering e2e tests need none of this to pass, despite using the same functionality. This needs to be investigated

louise-davies:
@sam-glendenning could perhaps using `initialData` and the query cache work for getting the count (see https://react-query-v3.tanstack.com/guides/initial-query-data#initial-data-from-cache)? If this doesn't work, then if we know that `incomingRecordCount` will already be calculated before any use of `recordCount`, perhaps we remove the `recordCount` query, and instead just populate `recordCount` from the query cache

louise-davies:
Also, good spot on the search vs filters issue. I don't think it's unreasonable behaviour to enforce that users need to click `Search`/`Apply` in order for their changes to be reflected so I'm not sure whether it is a "bug" as such that needs fixing.

sam-glendenning:
> @sam-glendenning could perhaps using `initialData` and the query cache work for getting the count (see https://react-query-v3.tanstack.com/guides/initial-query-data#initial-data-from-cache)? If this doesn't work, then if we know that `incomingRecordCount` will already be calculated before any use of `recordCount`, perhaps we remove the `recordCount` query, and instead just populate `recordCount` from the query cache

Ah good, this was the sort of thing I was looking for. I'll have a go at this

louise-davies (commented):
Did you manage to have a look at using the query cache? Submitting this as a review so it stops pinging me :)

sam-glendenning:
Sorry, I'd forgotten I'd committed to doing something more here. I'll mark it as draft until I need further review

sam-glendenning:
@louise-davies I've implemented your solution to reuse the cached data from `useIncomingRecordCount` in `useRecordCount` and it works well, nice fix!

I'd recommend giving this another review but this time in the context of how it runs in the browser. The most glaring problem at the moment is that the tooltip warning displays on searches we've already done (this can also mean that the user can perfrm a large search, get the results, then hit search over and over without changing any search parameters and the tooltip will display over and over).

Is there any way we can tell if a cached result is being used and therefore not show the tooltip if this happens? I am aware of an `isPreviousData` boolean returned from useQuery but I don't think this quite lines up with what I'm after because that only seems to compare the last query we ran, not all of them (and even then I couldn't seem to get it working quite right)

louise-davies:
@sam-glendenning I believe I have a solution to this using `queryClient.getQueriesData` to check if the `records` query has data for the given filters/search params combo the user is attempting to search. See gif below to see if you're happy with the new behaviour: (in the demo, I search with min shot number 1, change to another param and prompt the message without searching, switch back to 1 and click search without the message, and then go back to 2 and see the message again)

OG-search-warning

sam-glendenning:
That looks like what I'm after! Would you be able to commit that? I can write the tests, fix merge conflicts etc.

louise-davies:
@sam-glendenning committed :)

sam-glendenning:
@louise-davies I've got this ready for another review. The main functionality I was trying to test was:
1. initiate a search from date X and check it displays tooltip first
2. initiate a search from date Y and check it displays tooltip first
3. initiate a search from date X and check no tooltip displays, instead launching the search straight away (since the user has already seen the tooltip for this search)

Attempts to test this functionality in RTL failed (I was trying to `setQueryData` on a test query client but it didn't seem to pick up on any existing query data during the test). However, I did manage to test this in an e2e test so I would say this is adequate. If you want to unit test this, you can check what I committed in https://github.com/ral-facilities/operationsgateway/pull/149/commits/4d3fc61e3e67ee6dab5f1e5bbefb579fc33541bf which I have since reverted since it's no longer relevant.

cypress/integration/table/search.spec.ts
```diff
@@ -199,130 +455,164 @@ describe('Search', () => {
...
-        const timestampRange = condition['metadata.timestamp'];
+      cy.contains(/^Search$/).should('be.visible');
+    });
+  });
 
-        const gte: string = timestampRange['$gte'];
-        const lte: string = timestampRange['$lte'];
-        expect(gte).equal(expectedFromDateString);
-        expect(lte).equal(expectedToDateString);
+  describe.only('with record limits', () => {
```
louise-davies: Remove this only

louise-davies (approved):
I've just added/fixed tests from Sam's stuff, and I reviewed Sam's stuff before Christmas and it was all working good so I'll approve

PR #150, opened 2022-12-07
DSEGOG-199 DSEGOG-198 allow config of right y scale & min/max

louise-davies:
## Description
This PR adds the ability to independently change the scales & min/max of the right & left y axis.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-199 DSEGOG-198


src/plotting/plotSettings/plotSettingsController.component.tsx
```diff
@@ -61,8 +61,10 @@ export interface PlotSettingsControllerProps {
   changeXAxis: (value: string) => void;
   XAxisScale: XAxisScale;
   changeXAxisScale: (XAxisScale: XAxisScale) => void;
-  YAxesScale: YAxesScale;
-  changeYAxesScale: (YAxesScale: YAxesScale) => void;
+  leftYAxisScale: YAxesScale;
```
sam-glendenning: At one point, this data type was going to hold the scale of both y-axes but as we've gone in a different direction I'm in favour of renaming this to `YAxisScale`, similar to `XAxisScale`

sam-glendenning:
On my monitor this looks kinda cluttered. 
image

I'd be keen on having the scale options one on top of the other, I think that could look better, although I'm not sure what you've experimented with. So basically like the below, but one on top of the other (and the text saying left/right axis scale needing to be smaller to fit obviously)
image

src/plotting/plotSettings/yAxisTab.component.tsx
```diff
@@ -211,30 +236,95 @@ const YAxisTab = (props: YAxisTabProps) => {
...
+            </FormLabel>
+            <RadioGroup
+              row
+              aria-labelledby={`${
+                rightYAxisActive ? 'right' : 'left'
+              }-y-scale-group-label`}
+              name={`${
+                rightYAxisActive ? 'right' : 'left'
+              } y scale radio buttons group`}
+              value={rightYAxisActive ? rightYAxisScale : leftYAxisScale}
```
sam-glendenning: Do we need to split hairs over these labels and names being correct when we're only dealing with one y-axis?

sam-glendenning: On further inspection, I can see the need for retaining these values when respective y-axes are removed from the plot and then added back later so I'll resolve this one

src/plotting/plotWindow.component.tsx
```diff
@@ -59,8 +59,11 @@ const PlotWindow = (props: PlotWindowProps) => {
   const [XAxisScale, setXAxisScale] = React.useState<XAxisScale>(
     plotConfig.XAxisScale
   );
-  const [YAxesScale, setYAxesScale] = React.useState<YAxesScale>(
-    plotConfig.YAxesScale
+  const [leftYAxisScale, setLeftYAxisScale] = React.useState<YAxesScale>(
+    plotConfig.leftYAxisScale
+  );
+  const [rightYAxisScale, setrightYAxisScale] = React.useState<YAxesScale>(
```
sam-glendenning: Minor typo, for readability make this `setRightYAxisScale`

louise-davies:
> On my monitor this looks kinda cluttered. image
> 
> I'd be keen on having the scale options one on top of the other, I think that could look better, although I'm not sure what you've experimented with. So basically like the below, but one on top of the other (and the text saying left/right axis scale needing to be smaller to fit obviously) image

@sam-glendenning I was trying to limit the amount of vertical space that's taken up, since with users adding multiple channels vertical space is a bit of a luxury. (and the same thing needs to happen again for separate right and left y axis limits). @mattjSTFC thoughts? How should we lay out the scale checkboxes when the user has two y axis which need separate controls?

sam-glendenning:
It's true, but even something to separate the options out a bit could help. Perhaps a small bit of extra space between the two scales options in the middle?

louise-davies:
@sam-glendenning do you have any ideas for a shorter label names perhaps? Half the problem is that `left axis scale`/`right axis scale` are quite wordy labels 😅 I could make the font size smaller as well, which will give me leeway to put more space between the two

sam-glendenning:
Not really, short of a redesign where we do something like

```
Left Axis Scale (Linear O Log 0)
Right Axis Scale (Linear 0 Log O)
```

Theoretically that doesn't need to take up more vertical space than what's already there. Matt's input would be valuable though.

louise-davies:
I'll explore both - by default doing it vertically takes more space due to default MUI spacing

mattjSTFC:
Would something like this be a solution and covers either having the left axis only set or both:

image

image

The channel that is selected for either left or right is then displayed within the correct axis and not shown in one list under the search all channels field. Essentially behaves like a toggle.

I have set the left and right options as a radio button but if this doesn't stand out very well then is there an option to use the Toggle Button: https://mui.com/material-ui/react-toggle-button/#color

I may have missed some vital info here so let me know and I can revisit the wires.

sam-glendenning:
That could potentially work, @louise-davies what are your thoughts on this?

louise-davies:
That looks good, I think the radios don't stand out enough so a toggle button looks better imo, or we could use a [Switch](https://mui.com/material-ui/react-switch/) (specifically there's a custom example with labels on both sides)

I'll get to work on the functionality and I may give screenshots of all three options of radios, toggle buttons and a switch

sam-glendenning:
I am also in favour of a switch, was going to comment that but couldn't think of the name for it!

mattjSTFC:
I would not advise using a switch as I don't think this situation would expect one to appear. From NNG:

"Toggle switches are digital on/off switches. They prompt users to choose between two mutually exclusive options and always have a default value. Toggles should provide immediate results, giving users the freedom to control their preferences as needed."

https://www.nngroup.com/articles/toggle-switch-guidelines/

The options are related so I would go with a toggle button. An example, for a switch, would be a data cap on your mobile plan. It's either on or off.

louise-davies:
Cool, that makes sense, I'll go with a Toggle Button

louise-davies:
Here's a screenshot of the new layout:

image

sam-glendenning:
The Playwright tests continue to hound me too, to the point where I've set up an issue to track it (210). For now, all you can do is keep retrying until they pass

sam-glendenning (approved):
Seems to work nicely!

PR #153, opened 2022-12-09
DSEGOG-207 DSEGOG-209 frontend authorisation and amending timestamp format

sam-glendenning:
## Description
This adds the capability of authenticating with the backend and establishing a session using a JWT.

The latest version of the API also comes with an amendment to the format of datetime strings it expects. This PR adds a capital T between the date yyyy-MM-dd and the time HH:mm:ss and thus addresses DSEGOG-209. 

## Testing instructions

- [x] Review code
- [x] Check Actions build

## Agile board tracking
Closes DSEGOG-207 DSEGOG-209


src/index.tsx
```diff
@@ -176,6 +176,23 @@ if (
...
+    settings.then((settingsResult) => {
+      if (settingsResult) {
+        const apiUrl = settingsResult.apiUrl;
+        axios
+          .post(`${apiUrl}/login`, {
+            username: 'frontend',
+            password: 'front',
+          })
+          .then((response) => {
+            window.localStorage.setItem(MicroFrontendToken, response.data);
```
sam-glendenning, line 190: It is worth noting that this differs from DG plugins, which consists of:
```
const jwtHeader = { alg: 'HS256', typ: 'JWT' };
const payload = {
    sessionId: response.data.sessionID,
    username: 'Thomas409',
};
const jwt = jsrsasign.KJUR.jws.JWS.sign(
    'HS256',
    jwtHeader,
    payload,
    'shh'
);

window.localStorage.setItem(MicroFrontendToken, jwt);
```
Is the above still necessary here?

louise-davies: No. Previously, we needed to do that as we logged into ICAT, which gave us a session ID - but we need to have a JWT in DG. So in dev we sign our own JWT and put the ICAT info in it, which is normally scigateway-auth's job. Now, since in dev we're logging into og-api which returns a JWT, we can just use that.

src/parseTokens.ts
```diff
@@ -0,0 +1,30 @@
...
+      const code = p.charCodeAt(0).toString(16).toUpperCase();
+      return '%' + ('00' + code).slice(-2);
+    })
+  );
+  return payload;
+};
+
+type SciGatewayToken = string | null;
+
+export const readSciGatewayToken = (): SciGatewayToken => {
```
louise-davies, line 21: I presume this function is just to check we have a valid token (i.e. that we have a username) since we still use the underlying token in the Auth header

louise-davies (approved):
Works well, I can see that the API requires auth and our requests are succeeding

src/parseTokens.ts
sam-glendenning: I took this from DG and I think it's still a good idea for it to validate the token is good whenever the token is used (i.e. for requests and such). As it's stored in localStorage, technically the user could modify it in some way after all. But yeah it just checks we can still get a username out of it, which is somewhat robust but I suppose the ultimate check really is whether the backend accepts it or not. So we don't have to have the JWT parsing here if you don't think it's important I suppose

louise-davies: It's fine for now, it's potentially useful if info gets put in the JWT later

PR #158, opened 2022-12-12
DSEGOG-33 Upgrading to Cypress v12

sam-glendenning:
## Description
Updates from current version v9.7.0.

## Testing instructions
v9.7.0 should be gone and v12.0.2 in its place.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-33


sam-glendenning:
@louise-davies I know this was an issue assigned to you but I thought I'd tackle it since I had a bit of time before lunch today. It hasn't gone exactly to plan as you can see from the error message reported by Cypress.

image

I have looked through a variety of threads to try and understand the problem here and come up empty-handed when trying various solutions (plus I'd rather move onto another issue since this upgrade isn't too pressing) so I thought I'd leave this PR here for you to pick up if you want to build upon it if you find a fix (or I can finish what I've started).

I will link some threads I looked through below in case they are of use. Because I didn't see this error appear on the upgrade to Cypress 10 and beyond in SciGateway and DataGateway, I suspect Yarn plug-n-play and/or zero install are involved in this in some way and Cypress can't quite find what it's looking for (whether that be Typescript, the Cypress module itself or something else). I'm also fairly sure that the provided error message is closer to a red herring than an indication of what the actual problem might be.

https://stackoverflow.com/questions/73221655/unknown-file-extension-ts-cypress-config-ts-on-ci
https://github.com/cypress-io/cypress/issues/22096
https://github.com/cypress-io/cypress/issues/22747

sam-glendenning:
Also the base branch for this is currently the auth branch (for testing purposes) but I never got that far so we'll have to make sure it's changed back to develop after that's merged

louise-davies:
@sam-glendenning looks like the last issue you referenced seems like the likely culprit as they admit that `TS based cypress config` is not working with PnP. Since the upgrade isn't urgent, I will subscribe to that issue to keep tabs on it and look back into this once it's resolved

louise-davies (approved):
Works both locally and on CI - just had to use a specific version that works nicely with Yarn PnP and not use a ts config file and instead use a CJS js config file

PR #163, opened 2022-12-15
DSEGOG-213 Adding @yarnpkg/sdks dependency for automatic SDKs installation

sam-glendenning:
## Description
I was getting annoyed at having to run an extra command to get set up with the required SDKs for developing OperationsGateway while using Yarn 3 with PnP. We've got this documented in the repo wiki, which requires the user to run something like
```
yarn dlx @yarnpkg/sdks vscode
```
if they want to be able to resolve external dependencies in VSCode, for example. This is easily missed; I've done it myself several times and I wrote the damn thing.

There is a solution to make this happen automatically and this PR provides it.

First, we can add @yarnpkg/sdks to our `devDependencies`. What this does is it means we now only need to run the command `yarn sdks` and yarn will automatically install the correct SDKs for our environment. Another advantage of this is that renovate can now keep our SDKs up-to-date (in theory).

Second, now that we have this more generic command, we can add this to the `yarn postinstall` command to have it automatically run whenever the user runs `yarn install`. More users are aware of and should know to run `yarn install` before running `yarn start` so this improves the process overall.

## Testing instructions
The only way to properly test this is to clone a fresh copy of the repo in a brand new place on your computer. This eliminates the chance of any up-to-date lingering SDKs or Typescript libraries being resolved to the new workspace. Now open the workspace in your IDE of choice. If you're like me, the workspace will resolve the global version of Typescript I have installed on my computer (4.6.4 when we want the latest 4.9.3) and you won't be able to resolve external dependencies in the IDE.

Now run `yarn install`. This should set up the workspace with its normal dependencies plus its additional SDK files. Following this command's execution, your IDE _should_ now be able to resolve its external dependencies. If you're running VSCode, you should get a popup asking you to switch to the workspace version of Typescript (the newly installed 4.9.3), otherwise I believe this also happens automatically.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-213


sam-glendenning:
I am still getting a slight side-effect of running the `yarn sdks` command. The newly generated files are marked as modified by git even though it sees no content changes. Minor annoyance but I'd like to resolve it.
image
Googled the issue and many people say this is my Windows environment changing the line endings on those newly generated files to CRLF. To avoid this, I've enforced LF line endings in a new .gitattributes file but this doesn't seem to have worked.

@louise-davies can you check you get a similar thing on your machine (YMMV since you're running WSL). If not, it's probably just a local config thing I have set and it's not an issue.

louise-davies:
@sam-glendenning I can confirm that the overall process worked (i.e. cloned a new repo with this branch, opened VSCode in it, SDKs weren't working, ran `yarn install` and they work & VSCode prompted to use workspace VSCode version etc.). If I run `yarn sdks` I get no git changes (as expected in WSL). I'll test it in windows to see if your fix works on my machine

louise-davies:
@sam-glendenning does adding a `.editorconfig` file with the following contents fix it for you?

```
[*]
end_of_line = LF
charset = utf-8
```

It fixed things for me - initially I ran `yarn sdks` on windows and git had the files as changed. When I added that file it fixed things - to the point even if I remove the file I get no changes upon running `yarn sdks`. I supposed VSCode changed its settings when it was there... We only really need the `end_of_line` bit but on seeing examples of `.editorconfig` files setting the `charset` wouldn't hurt either

sam-glendenning:
No. But since these problems appear to be local to me, I'm not too worried about it affecting others. Are you happy for this to go in?

louise-davies (approved):

PR #164, opened 2022-12-15
DSEGOG-97 highlight search boxes when they automatically update

sam-glendenning:
## Description
This will need to be expanded upon when we start to incorporate experiments as a concept into the app but this lays the groundwork for it. 

I've reused the flash animation that appears on a `Chip` in the `filterInput` component when a user clicks the filter icon on a filtered table column.

1. When the user selects a timeframe range (and the datetime search box is populated with the relevant timestamps), the datetime search box displays the flashing animation
2. When a timeframe range is currently selected and the user edits either of the datetime fields manually (and the timeframe range box is reset back to a null value), the timeframe range box displays the flashing animation

## Testing instructions
Try steps 1 and 2 above. You should see the animations play.

- [x] Review code
- [x] Check Actions build
- [ ] {more steps here}

## Agile board tracking
(Partially) Closes DSEGOG-97


sam-glendenning:
Actually it might be better to say this _partially_ closes the issue, so we don't forget about the rest of the work required when experiments automatically update fields as well

louise-davies (commented):
The logic works well - the animation can get a bit janky if the user is quickly changing params (e.g. in the timeframe window)

sam-glendenning:
Yeah I had noticed that, I think it's just the way the new `sx` props are appended and removed from the element. I'll have another go on Monday to see if I can force through an improvement on it and update here

sam-glendenning:
I had a go at this today using the method Louise suggested. She suggested trying to clear the timeout that is set whenever the animation begins playing. This partly worked (in that I could clear the timeout) but I seemed to run into problems when trying to change the flag that actually stops the animation from playing and then immediately changing it back to true to get the animation to play again (so therefore the animation props never properly unmounted in the first place).

That was as far as I could get. It's worth having another go at this, it's a bit tricky when you're trying to change state manually and you've also got at least one timeout which is going to do it automatically as well, so you've got two things happening at once. But give it a go. Otherwise, what's here does a semi-decent job as it is.

louise-davies (approved):
What's here is good and partially closes the issue and lays the groundwork for the rest of the issue

PR #175, opened 2023-01-09
DSEGOG-200 hide y axes if no channels visible on them

louise-davies:
## Description
Fixes bug described in DSEGOG-200 - we now hide either the left or right Y axis depending on whether the user has made all the channels on that axis hidden. This works for both our own visibility mechanism as well as the native Chart.js mechanism

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Check that bug is fixed

## Agile board tracking
Closes DSEGOG-200


kennethnym (approved):
LGTM!

PR #176, opened 2023-01-18
DSEGOG-191 update plotting UI to separate time on x from channel on x plots

louise-davies:
## Description
Based on Stephen's feedback that timeseries plots are a separate "type" of plot to arbitary channel vs channel plots. A basic mock of this new UI was sent to and approved by Stephen.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-191


kennethnym:
I played around with the new separate time series plot and X/Y plot. The UI works well overall, but I found a potentially confusing interaction when switching between the 2 plotting modes. This is a video demonstrating the interaction:

https://user-images.githubusercontent.com/19359984/214026724-695492bf-ea94-4675-aa74-1b16773bc4ca.mp4

If I select `timestamp` for the X-axis in XY plotting mode, and then switch to time series mode, nothing happens as expected. However, when I switch back to XY plotting mode, the timestamp channel is erased. In fact, if I select any channel for the X-axis in XY mode, switch to time series mode, and then switch back, the selected channel is lost. This issue doesn't occur for Y-axis, as Y-axis channels are preserved when switching between modes.

- Is this intended behavior?
- Since the time series plot is now a separate mode, I think it makes sense to filter out the timestamp channel in XY mode

louise-davies:
@kennethnym 
>In fact, if I select any channel for the X-axis in XY mode, switch to time series mode, and then switch back, the selected channel is lost.

This is "intended" behaviour as the user is "selecting" a new X axis when switching to "Timeseries" mode - "timestamp". Otherwise we'd have to store the XAxis the user had selected in XY mode separately, and I don't think that's worth it for the complexity if the users don't explicitly require it.

I should have filtered timestamp from the dropdown for the XY mode though 😅 I'll fix that. I believe that will resolve the issues you showed in the demo.

kennethnym:
Got it. I will start the code review once you have pushed the changes.

src/plotting/plotSettings/xAxisTab.component.tsx
```diff
@@ -273,77 +274,81 @@ const XAxisTab = (props: XAxisTabProps) => {
...
-              InputLabelProps={{ style: { fontSize: 12 } }}
-              InputProps={{
-                ...params.InputProps,
-                style: { fontSize: 12 },
-                startAdornment: (
-                  <InputAdornment position="start">
-                    <Search />
-                  </InputAdornment>
-                ),
+      {XAxis !== 'timestamp' && (
```
kennethnym: I think it is a good idea to extract `'timestamp'` to a constant, because the value is referenced in different places, e.g. line 312 in this file, `plotSettingsController`, `chartTypeButtons`, etc.

src/plotting/plotSettings/xAxisTab.component.tsx
```diff
@@ -273,77 +274,81 @@ const XAxisTab = (props: XAxisTabProps) => {
...
+          </Grid>
+          <Grid container item>
+            <Autocomplete
+              disablePortal
+              freeSolo
+              clearOnBlur
+              id="select x axis"
+              options={allChannels
+                // don't let the user select timestamp in an XY plot
+                .filter((channel) => channel.systemName !== 'timestamp')
```
kennethnym: As mentioned in my previous comment.

src/plotting/plotSettings/xAxisTab.component.tsx
```diff
@@ -273,77 +274,81 @@ const XAxisTab = (props: XAxisTabProps) => {
...
+              )}
             />
-          )}
-        />
-      </Grid>
-      {XAxis && (
+          </Grid>
+        </>
+      )}
+      {XAxis && XAxis !== 'timestamp' && (
```
kennethnym: Prefer constant, as mentioned in my previous comment.

src/plotting/plotSettings/plotSettingsController.component.tsx
```diff
@@ -132,6 +132,59 @@ const PlotSettingsController = (props: PlotSettingsControllerProps) => {
     [setXYTabValue]
   );
 
+  const handleXAxisChange = React.useCallback(
+    (value?: string) => {
+      changeXAxis(value);
+      if (value === 'timestamp') {
```
kennethnym: Prefer constant, as mentioned in my comment on `xAxisTab.component.tsx` file.

src/plotting/plotSettings/plotSettingsController.component.tsx
```diff
@@ -141,57 +194,51 @@ const PlotSettingsController = (props: PlotSettingsControllerProps) => {
...
+        />
       </Grid>
       <Grid item>
-        <Box sx={{ borderBottom: 1, borderColor: 'divider' }}>
-          <Tabs
-            value={XYTabValue}
-            onChange={handleXYTabChange}
-            aria-label="tabs"
-            sx={{ height: 30, minHeight: 30 }}
+        {XAxis !== 'timestamp' && (
```
kennethnym: Prefer constant, as mentioned in my comment on xAxisTab.component.tsx file

src/plotting/plotSettings/plotSettingsController.component.tsx
```diff
@@ -141,57 +194,51 @@ const PlotSettingsController = (props: PlotSettingsControllerProps) => {
...
-            initialRemainingColours={remainingColours}
-            changeSelectedColours={changeSelectedColours}
-            changeRemainingColours={changeRemainingColours}
-          />
-        </TabPanel>
+            <Typography>X Axis Config</Typography>
+            {XAxisConfig}
+          </Paper>
+        )}
+        {XAxis !== 'timestamp' ? (
```
kennethnym: Prefer constant, as mentioned in my comment on xAxisTab.component.tsx file

src/plotting/plotSettings/chartTypeButtons.component.tsx
```diff
@@ -18,20 +22,59 @@
...
-      <ToggleButton value="scatter" aria-label="scatter chart">
-        <ScatterPlot />
-      </ToggleButton>
-      <ToggleButton value="line" aria-label="line chart">
-        <ShowChart />
-      </ToggleButton>
-    </ToggleButtonGroup>
+    <Stack direction="row" spacing={1}>
+      <ToggleButtonGroup
+        value={XAxis === 'timestamp' ? 'timeseries' : 'xy'}
```
kennethnym: Prefer constant, as mentioned in my comment on xAxisTab.component.tsx file.

src/plotting/plotSettings/chartTypeButtons.component.tsx
```diff
@@ -18,20 +22,59 @@
...
+        aria-label="chart type"
+      >
+        <ToggleButton value="timeseries" sx={{ textTransform: 'none' }}>
+          Timeseries
+        </ToggleButton>
+        <ToggleButton value="xy" sx={{ textTransform: 'none' }}>
+          XY
+        </ToggleButton>
+      </ToggleButtonGroup>
+      {XAxis === 'timestamp' && (
```
kennethnym: Prefer constant, as mentioned in my comment on xAxisTab.component.tsx file

.eslintrc.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
       files: ['e2e/**/*'],
       rules: {
         'testing-library/prefer-screen-queries': 'off',
+        'testing-library/no-await-sync-query': 'off',
```
kennethnym, line 40: Why is this eslint rule disabled?

kennethnym (changes requested):
Just some minor comments :)

.eslintrc.js
louise-davies: Because it's applying the wrong rules in playwright tests. Playwright has its own implementation of "RTL"-like selectors and so certain RTL rules are wrong e.g. in playwright all locators are async so we correctly await them, but this eslint rule is detecting that in RTL `getBy*` are sync functions and so we shouldn't await, despite that these are actually playwright locators (see https://playwright.dev/docs/locators)

We have to do similar for some cypress rules as well as you can see for similar reasons

kennethnym: Is it worth using the "overrides" settings in eslintrc to turn off the rule only for the e2e folder? Alternatively, an `.eslintrc` can be placed in the e2e folder that disables the offending rule only for that folder.

louise-davies: This is in the overrides section? And is just turning it off for files in the e2e folder?

kennethnym: Ahh yeah, you are right. How did I miss that? Sorry about that.

src/state/slices/tableSlice.tsx
```diff
@@ -38,7 +43,7 @@ export const tableSlice = createSlice({
     // Use the PayloadAction type to declare the contents of `action.payload`
     selectColumn: (state, action: PayloadAction<string>) => {
       // if it's the timestamp column, add to the beginning of the array
-      if (action.payload === 'timestamp') {
+      if (action.payload === timeChannelName) {
         state.selectedColumnIds.unshift(action.payload);
```
kennethnym, line 47: Is it me or is unshift a terrible name for what it does?

louise-davies: It is, hence why I added the comment to make it clear that that's what this is doing 😅

src/plotting/plotSettings/plotSettingsController.component.test.tsx
```diff
@@ -117,17 +68,91 @@ describe('Plot Settings component', () => {
...
-
-  it('renders plot settings form correctly (y-axis tab selected)', async () => {
-    const view = createView();
-    await user.click(screen.getByRole('tab', { name: 'Y' }));
-
-    expect(view.asFragment()).toMatchSnapshot();
+  describe('snapshots', () => {
+    beforeAll(() => {
+      jest.resetModules();
+      jest.doMock('./plotTitleField.component', () => (props) => (
```
kennethnym, line 74: Just curious - where are the `mock-*` components coming from? Is it a specific `jest.doMock` syntax that I am unaware of?

src/state/slices/tableSlice.tsx
kennethnym: If it wasn't for the comment I would have no idea what it does without googling.

kennethnym (approved):
LGTM!

src/plotting/plotSettings/plotSettingsController.component.test.tsx
louise-davies: The mock components are so that we can snapshot test the plotSettingsController without snapshotting all the underlying components. The tests therefore check that the correct props are being passed to the respective components. We're using `doMock` instead of `mock` so that the mocked components are only mocked in the snapshot tests, and in the other tests we're using the real components (and so can interact with them). The `mock-*` components are just nonsense JSX essentially.

PR #178, opened 2023-01-19
DSEGOG-137 data channel tree

louise-davies:
## Description
This PR adds the basic data channel dialogue & the tree view structure and removes the old channel list that was just for dev testing

Here's a gif demoing the new functionality

OG-channel-selecting-final

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-137


src/api/records.tsx
```diff
@@ -40,12 +41,7 @@ const fetchRecords = async (
   for (const [key, value] of Object.entries(sort)) {
     // API recognises sort values as metadata.key or channel.key
     // Therefore, we must construct the appropriate parameter
-    const sortKey = [
-      'timestamp',
-      'shotnum',
-      'activeArea',
-      'activeExperiment',
-    ].includes(key)
+    const sortKey = Object.keys(staticChannels).includes(key)
```
kennethnym: Since `staticChannels` is an object, and we are checking whether `key` exists as a key in `staticChannels`, I think `key in staticChannels` should also work.

src/channels/channelsDialogue.component.tsx
```diff
@@ -0,0 +1,192 @@
...
+            label="Search data channels"
+            fullWidth
+            InputProps={{
+              startAdornment: (
+                <InputAdornment position="start">
+                  <Search />
+                </InputAdornment>
+              ),
+            }}
+          ></TextField>
```
kennethnym: Tag can be ended with `/>` here.

src/channels/channelsDialogue.component.tsx
```diff
@@ -0,0 +1,192 @@
...
+        const node = prev.children[curr] as TreeNode;
+        // last item, so therefore is the channel
+        if (currIndex === splitPath.length - 1) {
+          node.children[channel.systemName] = {
+            ...channel,
+            checked: selectedIds.includes(channel.systemName),
+          };
+        }
+        // return the child node, so we drill deeper into the tree
+        return node;
```
kennethnym, line 81: I really like this piece of code here :)

src/app.types.tsx
```diff
@@ -53,31 +53,46 @@ export type DataType = 'scalar' | 'image' | 'waveform';
...
 export type FullChannelMetadata =
   | FullScalarChannelMetadata
   | FullImageChannelMetadata
   | FullWaveformChannelMetadata;
 
+// Type guards because TS can't deal with nested discriminated unions
+export const isChannelMetadataScalar = (
+  c: FullChannelMetadata
+): c is FullScalarChannelMetadata => c.type === 'scalar';
+export const isChannelMetadataImage = (
```
kennethnym, line 89: This seems to be unused. Will this be utilized in the future?

src/filtering/filterParser.ts
```diff
@@ -128,9 +130,7 @@ const convertOperator = (opToken: Token): string => {
  * convertChannel("CHANNEL_1")
  */
 const convertChannel = (channel: string): string => {
-  if (
-    ['timestamp', 'shotnum', 'activeArea', 'activeExperiment'].includes(channel)
-  ) {
+  if (Object.keys(staticChannels).includes(channel)) {
```
kennethnym: I think `channel in staticChannels` should also work here as well.

kennethnym (changes requested):
The UI works well, although I think something else can be used instead of a disabled checkbox to indicate "there are subitems in this item" in the tree view. For me, it is hard to tell between a selectable channel vs a tree node that contains children, because the checkboxes look the same besides the *slightly* fainter color of the disabled ones.

I think an MUI "arrow-left" or "angle-left" icon can be used instead to indicate non-leaf nodes.

louise-davies:
@kennethnym There is actually a later issue to address this kinda: https://stfc.atlassian.net/browse/DSEGOG-140 - I didn't want to add more stuff to this PR as it's already quite big. That's the reason the checkboxes are rendered at all - I'll probably experiment with different ways of indicating when addressing the later issue - so I'd like this PR review to ignore the checkboxes for non-leaf nodes as this is something to be changed later

src/api/records.tsx
louise-davies: Oops, I think because I as replacing an array I was in "change object into array" rather than "refactor this check" mode 😅

src/app.types.tsx
louise-davies: It likely will be - I just created TS type assertion functions for all three since they're all related. We haven't had to assert specifically that a channel/it's metadata is an image yet, but likely will when we start doing image manipulation in a later milestone

src/api/records.tsx
kennethnym: Haha, it happens!

kennethnym (approved):
Got it. LGTM!

PR #189, opened 2023-02-02
DSEGOG-138 DSEGOG-139 Add channel metadata panel

louise-davies:
## Description
This PR fills in the channel metadata panel with both channel metadata and the data summary. Note, the channel summary endpoint currently mismatches what the API returns (it will be changed soon), hence the addition of the `msw` library to enable easy mocking of requests (TODO use this for all testing - I'll create an issue to explore this as this would be very cool to have centralised "test" endpoints)

EDIT: I've now refactored the e2e tests to use `msw` - let me know thoughts. Essentially, it makes "happy path" mocking easier i.e. if we just want what our default handlers call, there's no config. However, it makes it more difficult to inspect the requests (as seen in the Cypress tests). And replacing a mock for an individual test is about the same as what we were doing previously

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-138, Closes DSEGOG-139


kennethnym:
Very excited about the addition of msw - might be worth exploring adopting it in unit tests in the future, so no more mock axios get, `if url == "/abc"` and `promise.resolve` in every test file

louise-davies:
@kennethnym yep, I'll create a separate task to explore using it in unit tests as it's a larger change and so I wouldn't want it bundled into a feature.  I kind of got carried away ensuring that I could use `msw` in the e2e tests as it is 😅 But yeah seems really promising, as well as enabling development when the server is down/when an endpoint doesn't exist yet

kennethnym:
Absolutely. I am honestly down to adopt msw in datagateway, especially since we are already in the process of migrating tests to rtl. Lots of tests now require mock axios get, so might as well do it in msw!

louise-davies:
Yeah, we could always just adopt it in jest first and later see about propogating it into cypress as well. I'd be down for it after seeing how it works in this repo. I had been meaning to check it out for OG for a while - but hadn't had a good enough reason to until this PR as there's an endpoint that needs to return different data from what the backend's currently returning and so to develop I needed to mock the response until they change it!

src/channels/channelMetadataPanel.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+              <Table aria-label="recent data" size="small">
+                <TableHead>
+                  <TableRow>
+                    <TableCell>Time</TableCell>
+                    <TableCell>Data</TableCell>
+                  </TableRow>
+                </TableHead>
+                <TableBody>
+                  {channelSummary.recent_sample.map((sample) => {
+                    const [timestamp, data] = Object.entries(sample)[0];
```
kennethnym, line 121: On my end, `sample` can sometimes be a number rather than an `Object`, causing this to crash.

For example, `channelSummary` for this data channel tree path: All Channels > LA3 > N_PUMP > IMAGE > North pump near field energy looks like this:

```json
{
    "first_date": "2022-04-07T14:16:16",
    "most_recent_date": "2022-04-08T16:58:57",
    "recent_sample": [
        15.046,
        15.066,
        14.979
    ]
}
```

src/channels/channelMetadataPanel.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    const formattedTimestamp = renderTimestamp(timestamp);
```
kennethnym, line 122: This sometimes also causes issue for me when `timestamp` is `'0'`. Can be reproduced following this path: All Channels > Dataset > Data type. Should be able to avoid this issue by doing a simple truthy check on the timestamp - if `timestamp` is falsy, return "unknown timestamp", otherwise format the timestamp.

src/api/records.tsx
```diff
@@ -217,22 +219,16 @@ export const useRecordsPaginated = (): UseQueryResult<
                 break;
               case 'image':
```
kennethnym: Consider using the type guards for `Channel` type provided [here](https://github.com/ral-facilities/operationsgateway/blob/develop/src/app.types.tsx#L86-L94).

```ts
import { isChannelImage, isChannelScalar, isChannelWaveform } from './app.types';

if (isChannelScalar(channel)) {
  // ...
} else if (isChannelImage(channel)) {
  // ...
} else if (isChannelWaveform(channel)) {
  // ...
}
```

src/api/records.tsx
```diff
@@ -217,22 +219,16 @@ export const useRecordsPaginated = (): UseQueryResult<
...
                 channelData = (channel as ImageChannel).thumbnail;
-                channelData = (
-                  <img
-                    src={`data:image/jpeg;base64,${channelData}`}
-                    alt={key}
-                    style={{ border: '1px solid #000000' }}
-                  />
+                channelData = renderImage(
```
kennethnym: Is it possible to move the rendering of data to `constructColumns` in `channels.tsx` [here](https://github.com/ral-facilities/operationsgateway/blob/develop/src/api/channels.tsx#L120-L128)? I am under the impression that the data passed to the `useTable` hook should only be pure data, and the presentation should be handled by the individual Columns through the `Column.Cell` option.

cypress/support/commands.js
```diff
@@ -71,3 +71,50 @@ Cypress.Commands.add('dragAndDrop', (subject, target) => {
         });
     });
 });
+
+Cypress.Commands.add('clearMocks', () => {
+  mockedRequests = [];
+});
+
+let mockedRequests = [];
```
kennethnym: Since `clearMocks` command also uses this variable, consider moving this before `clearMocks` for clarity's sake.

package.json
```diff
@@ -117,10 +120,14 @@
     "husky": "8.0.1",
     "jest-canvas-mock": "2.4.0",
     "lint-staged": "13.1.0",
+    "msw": "^1.0.0",
```
kennethnym: Should this be pinned like other dependencies?

kennethnym (changes requested):
Stumbled across some issues when running the code. Noted them down in the comments.

src/channels/channelMetadataPanel.component.tsx
louise-davies: Ah yeah, this is the changed endpoint I highlighted above - you need to use `msw` for this endpoint specifically. Here's the new example response from @MRichards99  - it just hasn't been changed or deployed in the backend yet!

> ```
> {
>     "first_date": "2022-04-07T14:16:16",
>     "most_recent_date": "2022-04-08T16:58:57",
>     "recent_sample": [
>         {"2022-04-08T14:16:16": -8582000.0},
>         {"2022-04-09T14:16:16": -8461000.0},
>         {"2022-04-10T14:16:16": -8535000.0}
>     ]
> }
> ```

src/channels/channelMetadataPanel.component.tsx
louise-davies: Urgh JS type coercion. I mean I wouldn't think we'd ever expect a timestamp of `'0'` but it's theoretically possible so....

package.json
louise-davies: I pin it in my next PR but I'll pin it here too for good measure 👍

src/api/records.tsx
louise-davies: Yup, this is older code from before the type guards - good catch

src/channels/channelMetadataPanel.component.tsx
kennethnym: Got it, code is working now after enabling msw.

src/api/channels.tsx
```diff
@@ -126,6 +172,14 @@
         ) : (
           <React.Fragment>{String(value ?? '')}</React.Fragment>
         );
+    } else {
```
kennethnym, line 175: Just as a note here, this should be fine assuming there are only 3 types of channel metadata and only scalar type is non-image.

src/api/records.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react';
 import axios, { AxiosError } from 'axios';
```
kennethnym, line 1: Since React import is removed, is it safe to rename the file extension to `.ts`? Same for `records.test.tsx`.

src/table/cellRenderers/cellContentRenderers.test.tsx
```diff
@@ -72,4 +76,33 @@ describe('cell content renderers', () => {
...
+          alt="alt text"
+          src="data:image/jpeg;base64,base64"
+          style={
+            Object {
+              "border": "1px solid #000000",
+            }
+          }
+        />
+      `);
+    });
```
kennethnym, line 94: Maybe a test case for empty/falsy base64 string should be added here?

kennethnym (changes requested):
Code looks good :) Just some minor changes.

src/api/channels.tsx
louise-davies: Yeah, this is correct - otherwise TS would complain - the minute we add another channel type/channel metadata type this would complain in TS. Both `waveform` and `image` are represented by thumbnails in the table

src/api/records.tsx
louise-davies: I might keep it as a `.tsx` file for a couple of reasons: 1. `channels.tsx` is a similar file and it makes sense for them to be named similarly and 2. it may be in the future we add more `select` functions that do generate JSX code and thus would require renaming back to `records.tsx`.

kennethnym: I see. In that case I would be inclined to keep it as is as well.

kennethnym (approved):
LGTM!

PR #193, opened 2023-02-07
DSEGOG-231 migrate unit tests to use `msw`

louise-davies:
## Description
Migrate unit tests to use `msw` library added in the channel metadata PR. Additionally, I also tidied up some errors that were being thrown in the unit tests, fixed some flaky tests caused by `user.type` being slow (so reduced the amount of characters we type!), as well as fixing a bug in the filter input caused by not using the system name as the React key

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-231


src/api/channels.test.tsx
```diff
@@ -286,24 +199,31 @@ describe('channels api functions', () => {
...
       const { result } = renderHook(() => useChannelSummary(undefined), {
         wrapper: hooksWrapperWithProviders(),
       });
 
       expect(result.current.isFetching).toBeFalsy();
       expect(result.current.isLoading).toBeTruthy();
-      expect(axios.get).not.toHaveBeenCalled();
+      expect(requestSent).toBe(false);
 
       const { result: result2 } = renderHook(
```
kennethnym: Consider using the `rerender` function returned by `renderHook` here. The `renderHook` callback accepts arbitrary props as the first argument. New props can be passed to the hook through `rerender(newProps)`:

```ts
import { useChannelSummary } from './channels';
import { hooksWrapperWithProviders } from './setupTests';

const { rerender } = renderHook((channel?: string) => useChannelSummary(channel), {
  wrapper: hooksWrapperWithProviders(),
  // this will be forwarded to the `channel` param of the callback
  initialProps: undefined,
})

// ...

// re-render hook with another channel name.
rerender('timestamp');
```

src/api/channels.test.tsx
```diff
@@ -314,7 +234,7 @@ describe('channels api functions', () => {
 
       expect(result2.current.isFetching).toBeFalsy();
       expect(result2.current.isLoading).toBeTruthy();
-      expect(axios.get).not.toHaveBeenCalled();
+      expect(requestSent).toBe(false);
     });
 
     it.todo(
```
kennethnym, line 239: Is this test obsolete, or is it still pending implementation in `msw`?

src/channels/channelsDialogue.component.test.tsx
```diff
@@ -43,32 +39,60 @@ describe('selectChannelTree', () => {
...
+              children: {
+                ...Object.fromEntries(
+                  testChannels
+                    .filter((channel) => channel.path.includes('1'))
+                    .map((channel) => [
+                      channel.systemName,
+                      {
+                        ...channel,
+                        checked:
+                          channel.systemName === 'CHANNEL_ABCDE' ? true : false,
```
kennethnym: Minor point, but this can be shortened to `channel.systemName === 'CHANNEL_ABCDE'`.

src/channels/channelsDialogue.component.test.tsx
```diff
@@ -43,32 +39,60 @@ describe('selectChannelTree', () => {
...
           checked: false,
           children: {
-            [testChannels[4].systemName]: {
-              ...testChannels[4],
-              checked: true,
+            '1': {
+              name: '1',
+              checked: false,
+              children: {
+                ...Object.fromEntries(
```
kennethnym: I think this can be combined with a `reduce` call on `testChannels` - check if `channel.path` includes `'1'`, if so, add channel to the object.

kennethnym (changes requested):
Nice refactor! Nice to see axios mocks going away. I left some comments below.

src/api/channels.test.tsx
louise-davies: They're still pending the fact that we haven't decided how we want to handle errors from the API in OG yet - you can see we're currently just logging errors right now - so keeping these tests as `todo`s reminds us that there's no proper API error handling yet

src/api/channels.test.tsx
louise-davies: Ah, I didn't realise that there's a `rerender` function from `renderHook` - good catch this simplifies the test code significantly :)

kennethnym (dismissed):
LGTM!

kennethnym (approved):
A stronger LGTM coming in hot!

PR #196, opened 2023-02-13
DSEGOG-135 Channel Search

louise-davies:
## Description
Add the ability to search for channels using a MUI autocomplete, which upon selecting a channel will show that channel in the metadata panel & show the correct place in the channel tree. Channels are searchable by both friendly names and system names.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-135

### DIFF
--- a/src/channels/channelsDialogue.component.test.tsx
+++ b/src/channels/channelsDialogue.component.test.tsx
@@ -183,4 +183,18 @@ describe('Channels Dialogue', () => {
     ]);
     expect(props.onClose).toHaveBeenCalled();
   });
+
+  it('selects a channel in the metadata panel & channel tree when search box is used', async () => {
+    createView();
+
+    const search = screen.getByLabelText('Search data channels');
+
+    await user.type(search, 'shot{arrowdown}{enter}');
+
+    // shot number should be visible in tree view
+    screen.getByRole('checkbox', { name: 'Shot Number' });
+
+    // shot number should be selected in the metadata panel
+    screen.getByText('System name: shotnum');
+  });
 });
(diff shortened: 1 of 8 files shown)


src/channels/channelsDialogue.component.test.tsx
```diff
@@ -183,4 +183,18 @@ describe('Channels Dialogue', () => {
...
+
+  it('selects a channel in the metadata panel & channel tree when search box is used', async () => {
+    createView();
+
+    const search = screen.getByLabelText('Search data channels');
+
+    await user.type(search, 'shot{arrowdown}{enter}');
+
+    // shot number should be visible in tree view
+    screen.getByRole('checkbox', { name: 'Shot Number' });
```
kennethnym: Should this be wrapped in an `expect` assertion?
```js
expect(screen.getByRole('checkbox', { name: 'Shot Number' })).toBeInTheDocument();
```

src/channels/channelsDialogue.component.test.tsx
```diff
@@ -183,4 +183,18 @@ describe('Channels Dialogue', () => {
...
+
+    // shot number should be selected in the metadata panel
+    screen.getByText('System name: shotnum');
```
kennethnym: Same as above.

kennethnym (changes requested):
Code looks good and works well, just one minor comment :)

src/channels/channelsDialogue.component.test.tsx
louise-davies: Technically it doesn't have to be, as `getByRole` would error if it wasn't in the document, but I can wrap in `expect` to make it more explicit

kennethnym: Yeah, I think wrapping it in an expect makes the intention clearer, especially for devs who might not be aware that `getBy*` queries throw when no element is found.

kennethnym (approved):
LGTM!

PR #215, opened 2023-03-14
DSEGOG-125 DSEGOG-126 DSEGOG-127 false colour

louise-davies:
## Description
Add a panel to the image window to adjust false colour parameters for images.

Also, realised we could fix some minor issues with the popup windows when the main window is hidden (either by minimisation of main window, maximisation of popup, changing tab in main window etc.) which were caused by browsers polling timers less frequently. Some issues that were fixed were tooltips entering/leaving in popups being delayed, and with the false colour work the loading of new images/colourbars was also delayed.

See gif demoing the behaviour:

OG false colour

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes DSEGOG-125, DSEGOG-127, DSEGOG-126


joshdimanteto (approved):
LGTM

PR #218, opened 2023-03-27
DSEGOG-227 Changes to datapicking and timeframes

joshdimanteto:
## Description
Remove the consideration of seconds when picking dates and setting timeframes.
Set the seconds of the from date to 00 and the seconds of the to date to 59.
Disable the selection of seconds in the datepicker.

## Testing instructions
- Verify that the timeframes are set correctly by ensuring that the from date has its seconds set to 00 and the to date has its seconds set to 59.
- Confirm that the datepicker cannot select seconds and that the from and to dates have their seconds set to 00 and 59, respectively. 

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
Closes DSEGOG-227


louise-davies (changes requested):
I think there's also the time pickers that are part of the plot settings that need changing as well (found in `xAxisTab.component.tsx`)

cypress/integration/table/search.spec.ts
```diff
@@ -54,7 +54,7 @@ describe('Search', () => {
         '2022-01-01 00:00:00'
       );
       cy.get('input[aria-label="to, date-time input"]').type(
-        '2022-01-02 00:00:00'
+        '2022-01-02 00:00:59'
```
louise-davies: Shouldn't we not type in the seconds? As users won't be typing in seconds

cypress/integration/table/search.spec.ts
```diff
@@ -646,7 +646,7 @@ describe('Search', () => {
...
```
louise-davies: Again, should probably just type `00:00` instead of specifying the seconds to mirror expected user behaviour

louise-davies (commented):
Minor comment about tests, but otherwise it all seems to be working correctly :)

cypress/integration/table/search.spec.ts
joshdimanteto: I've changed it in the unit and e2e tests in the latest commit

louise-davies (approved):
LGTM

PR #221, opened 2023-04-06
DSEGOG-92 DSEGOG-91 DSEGOG-90 DSEGOG-94 DSEGOG-90 DSEGOG-96 DSEGOG-97 searching for data using experiment id

joshdimanteto:
## Description
- Allow the experiment id to be searchable
- The experiment ID should update the to and from dates
- to and from dates should update the shot number and visa versa
- The Timeframe must be reset when an experiment ID is chosen and visa versa 
- The Experiment Id must be reset when a date is manually inputted into the to and from date input boxes 
- Add the flash animation to the other search box
- Add experiment shaded region on "to" date picker when a "from" date is selected and visa versa
- Search within experiment Timeframe using date picker and shot number (This doesn't work if you clear and retype in a date).  

## Testing instructions
- MSW mock and the test api url 

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-92, closes DSEGOG-91, closes DSEGOG-90, closes DSEGOG-94, closes DSEGOG-90, closes DSEGOG-96, closes DSEGOG-97


joshdimanteto:
experiment id

joshdimanteto:
shaded region
date-to-shotnum

joshdimanteto:
experiment id search functionailty

src/search/components/shotNumber.component.tsx
```diff
@@ -92,6 +105,31 @@ const ShotNumber = (props: ShotNumberProps): React.ReactElement => {
   const invalidRange =
     min !== undefined && max !== undefined ? min > max : false;
 
+  const [flashAnimationPlaying, setFlashAnimationPlaying] =
+    React.useState<boolean>(false);
+
+  // Stop the flash animation from playing after 1500ms
+  React.useEffect(() => {
+    if (
+      (!props.searchParameterShotnumMax && !props.searchParameterShotnumMin) ||
```
louise-davies: Need to use proper `typeof x !== undefined` checks here, as these falsy checks can trigger when min and max are 0 (which are valid values but causes the flash to happen)

src/search/components/shotNumber.component.tsx
```diff
@@ -92,6 +105,32 @@ const ShotNumber = (props: ShotNumberProps): React.ReactElement => {
...
     min !== undefined && max !== undefined ? min > max : false;
 
+  const [flashAnimationPlaying, setFlashAnimationPlaying] =
+    React.useState<boolean>(false);
+
+  // Stop the flash animation from playing after 1500ms
+  React.useEffect(() => {
+    if (
+      (typeof props.changeSearchParameterShotnumMax !== undefined &&
+        typeof props.changeSearchParameterShotnumMin !== undefined) ||
```
louise-davies: Can these ever be `undefined`? I thought these are functions? Previously we were doing `!props.searchParameterShotnumMax && !props.searchParameterShotnumMin`

joshdimanteto: Sorry, that's my mistake. I meant to put `props.searchParameterShotnumMax` and `props.searchParameterShotnumMin`.

joshdimanteto: I have fix it in the latest commit. I also noticed the logic was wrong as well  and I fix that as well.

louise-davies (changes requested):
I found a bug when testing the shotnumber range, and noticed some inconsistent behaviour. It clears the experiment ID the first time but if I repeat the steps then it doesn't clear it. Looking at the range I see that it is in range of the experiment and so should not clear. Perhaps because the value returned from range converter isn't defined on the first attempt and so experiment ID clears before it's fetched, but next time it's already fetched and so it doesn't clear? See GIF

OG-experiment-bug

joshdimanteto:
> I found a bug when testing the shotnumber range, and noticed some inconsistent behaviour. It clears the experiment ID the first time but if I repeat the steps then it doesn't clear it. Looking at the range I see that it is in range of the experiment and so should not clear. Perhaps because the value returned from range converter isn't defined on the first attempt and so experiment ID clears before it's fetched, but next time it's already fetched and so it doesn't clear? See GIF
> 
> OG-experiment-bug

Addressed in [514a1b8](https://github.com/ral-facilities/operationsgateway/pull/221/commits/514a1b86b70d88b144888d89f56f579efdee1fe9)

- The problem occurred because I wasn't using the dates received from the range convertor directly

joshdimanteto:
exper-id

joshdimanteto:
In the GIF above. I noticed that the date range animation wasn't triggering the second time around. I have addressed this in  [55f2361](https://github.com/ral-facilities/operationsgateway/pull/221/commits/55f2361a4fc20e4af99563d0bba25ad06b70dc8f)  

exper-id-2

louise-davies (approved):
The fixes you added look good, and playing around a little more reveals no further bugs that I can see so LGTM

PR #222, opened 2023-04-13
Add an auto refresh toggle when relative timeframe is set

kennethnym:
## Description
This PR adds a toggle next to the refresh button that allows users to enable auto refresh of data when a relative time frame is set, e.g. "last 10 minutes".

I set the auto refresh to happen every minute, but happy to change it if needed.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
connect to DSEGOG-86

### DIFF
--- a/src/search/components/__snapshots__/maxShots.component.test.tsx.snap
+++ b/src/search/components/__snapshots__/maxShots.component.test.tsx.snap
@@ -7,7 +7,7 @@ exports[`maxShots search renders correctly with a selected value 1`] = `
   >
     <div
       aria-label="select max shots to display"
-      class="MuiBox-root css-1hqcbhu"
+      class="MuiBox-root css-2sv9t8"
     >
       <div
         style="align-items: center; display: flex; justify-content: left; margin-right: 10px;"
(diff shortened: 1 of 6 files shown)


louise-davies (changes requested):
Code looks good, however when a user first loads up the app, the first time they select a relative timeframe the auto refresh toggle quickly flashes into view then away again. See GIF

OG-auto-refresh

Could you investigate this please? Also, I believe that the auto-refresh toggle should have the same "behaviour" as the refresh button itself - either both the buttom and the toggle should be hidden when a non-relative timeframe search is used, or both are visible but disabled. It's a bit odd to have one doing one and one doing the other. If we decide to just disable them both then that fies the issue in the GIF, but I wouldn't mind hiding them both if the GIF issue is fixed.

kennethnym:
I think I know why it is happening - the toggle's appearance depends on `timeframeRange` & `paramsUpdated`. When the user selects a time frame, `timeframeRange` is set immediately, but `paramsUpdated` at that time is still `false`, so the toggle briefly appears, then `paramsUpdated` is correctly updated in a `useEffect`, which then hides the toggle.

louise-davies (approved):
LGTM

louise-davies (approved):
LGTM

PR #225, opened 2023-05-02
 DSEGOG-247 using space to enter a token in filters

joshdimanteto:
## Description
- In operations gateway you can currently only build filters using enter. 
- Allow filters to be build using space   

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-247


src/filtering/filterInput.component.tsx
```diff
@@ -62,8 +62,74 @@ const FilterInput = (props: FilterInputProps) => {
...
+            value: inputValue,
+            label: inputValue,
+          });
+
+          setValue(newValue);
+          setInputValue('');
+          setError('');
+          setInputIndex((prevIndex) => prevIndex + 1);
+        }
+      }
```
louise-davies: There's a lot of repeated code here, whereas the only thing that's different in all these cases is what gets `splice`d. I'd change this to have a variable to store the new item `let newToken`, and have the `if` statements set the value of `newToken`, then at the end if `newToken` isn't undefined (i.e. one of the conditions matched and set `newToken` to something), then run the common code:

```
e.preventDefault();
e.stopPropagation();

newValue.splice(inputIndex, 0, newToken);

setValue(newValue);
setInputValue('');
setError('');
setInputIndex((prevIndex) => prevIndex + 1);
```

src/filtering/filterInput.component.tsx
```diff
@@ -62,8 +62,74 @@ const FilterInput = (props: FilterInputProps) => {
         setError('');
         setInputIndex((prevIndex) => prevIndex + -1);
       }
+      // allow selecting operators, channels, strings
+      // and numbers with Space key down
+      if (e.key === ' ') {
+        const operator = operators.find((opt) => opt.value === inputValue);
```
louise-davies: I wonder if it would be nice to have the same behaviour as the channel behaviour - if there's only 1 operator left as an option then space selects it. Useful for some operators like "not" where it saves you having to type the `o` and `t`

louise-davies (changes requested):

src/filtering/filterInput.component.tsx
joshdimanteto: I thought about that but the reason why i didn't implement it like that was because it wouldn't work for all operators such as "<"  because "<=" exists there would always be two values.

louise-davies: ah true - I thought that there was a reason but I couldn't remember what. You could handle the cases separately I suppose, i.e. where operator matches exactly even if there's more than one, and if there's only one as well, but that may be a little overkill 😅

src/filtering/filterInput.component.tsx
```diff
@@ -62,8 +62,82 @@ const FilterInput = (props: FilterInputProps) => {
...
+          (channel.length === 1 ||
+            operatorExactMatch !== undefined ||
+            operatorListMatch.length === 1 ||
+            (!Number.isNaN(Number(inputValue)) &&
+              inputValue.trim().length > 0) ||
+            (inputValue[0] === '"' &&
+              inputValue[inputValue.length - 1] === '"') ||
+            (inputValue[0] === "'" &&
+              inputValue[inputValue.length - 1] === "'")) &&
+          newToken
```
louise-davies: Can't we just test if `typeof newToken !== undefined`? As `newToken` by definition is only defined when one of the cases in the switch statement matches. Means we don't have to repeat the case match logic here

joshdimanteto: I've done the changes below. I made one change to use  `newToken !== undefined` as the type of newToken is `Token | undefined` and using `typeof newToken !== undefined` causes error as it `true` when `newToken` is `undefined`

louise-davies: Ah yeah, I made a typo - the typeof version of the check needs to be `typeof newToken !== "undefined"`. However, that check is just an extra safe version of `newToken !== undefined`, as the latter errors if `newToken` isn't defined, but we know in our code that `newToken` is always declared but perhaps not set. It's just good practice to use `typeof newToken !== "undefined"` if what you're checking can actually also be not be declared as a variable yet.

louise-davies (approved):
LGTM! Works well, essentially meaning any time you can press enter and it accepts the current token pressing space works too :)

src/filtering/filterInput.component.tsx
joshdimanteto: Thanks for the clarification. I have made the changes below.

PR #244, opened 2023-05-05
DSEGOG-252 Add additional error checks for shotnum ranges and date ranges

joshdimanteto:
## Description
- Add error state when one field is filled in for the dates and the shot numbers
- Remove the the shot number being send to the API as the range converter deal with this now  
- Disable the search button when only one field is filled in
- Modify the experiment start and ends dates to match the records (ps not all experiments fall within range of records)

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-252


louise-davies (approved):
LGTM

PR #248, opened 2023-05-16
DSEGOG-251 DSEGOG-252 Improvements to the false colour

joshdimanteto:
## Description
Enter a description of the changes here

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-251, closes DSEGOG-252


joshdimanteto:
false colour

e2e/images.spec.ts-snapshots/user-can-change-the-false-colour-to-use-reverse-1-chromium-linux.png
louise-davies, line 1: This screenshot doesn't match the others or the colourbar so it was probably taken too early I'd guess? Let's wait until Kenneth's playwright stuff is merged and fix it after that

e2e/traces.spec.ts-snapshots/user-can-change-trace-via-clicking-on-a-thumbnail-1-webkit-linux.png
louise-davies, line 1: This snapshot shouldn't have changes - it now no longer matches the other browsers. Again, let's wait for the playwright docker stuff

src/images/falseColourPanel.component.tsx
```diff
@@ -55,6 +63,111 @@ interface FalseColourPanelProps extends FalseColourParams {
   changeUpperLevel: (value: number | undefined) => void;
 }
 
+function MyListSubheader(props: ListSubheaderProps) {
```
louise-davies: What is the point of this custom component? Might as well just use `ListSubheader`

src/images/falseColourPanel.component.tsx
```diff
@@ -55,6 +63,111 @@ interface FalseColourPanelProps extends FalseColourParams {
...
+  );
+};
+
+const ColourMapSelectExtended = (
+  colourMap: string,
+  handleColourMapChange: (event: SelectChangeEvent) => void,
+  extendedColourMapNames: ColourMapsParams
+) => {
+  const mainColourMapTypeNames = Object.keys(extendedColourMapNames);
+  const mainColourMapNames = Object.values(extendedColourMapNames);
```
louise-davies: These variables should probably be renamed to have the prefix `extended` rather than `main`

src/images/falseColourPanel.component.tsx
```diff
@@ -55,6 +63,111 @@ interface FalseColourPanelProps extends FalseColourParams {
...
+          {colourMap}
+        </MenuItem>
+      ))}
+
+      <MyListSubheader>{mainColourMapTypeNames[6]}</MyListSubheader>
+      {mainColourMapNames[6].map((colourMap) => (
+        <MenuItem key={colourMap} value={colourMap}>
+          {colourMap}
+        </MenuItem>
+      ))}
```
louise-davies: Could this code not be achieved with a loop over `mainColourMapTypeNames`/`mainColourMapNames` to reduce code repetition? Or just a simple for loop, and access both arrays by index.

louise-davies: Also, if this is then done by a loop, then there wouldn't need to be two separate components for `ColourMapMain` and `ColourMapExtended` - the difference between them would just be that the main one is only supplied with one item to "loop" over

src/images/falseColourPanel.component.tsx
```diff
@@ -99,127 +208,175 @@ const FalseColourPanel = (props: FalseColourPanelProps) => {
...
+    if (selectColourMap !== '') {
+      if (!checked) {
+        changeColourMap(selectColourMap);
+      } else {
+        changeColourMap(`${selectColourMap}_r`);
+      }
+    }
+    setReverseColour(checked);
+  };
+  function filterNamesWithSuffixR(
```
louise-davies: This function can be extracted outside the component - a slight performance optimisation

src/images/falseColourPanel.component.tsx
joshdimanteto: The menuItems do not work correctly when rendered in a loop. When I initially did it a loop the menuItems were not clickable.

src/images/falseColourPanel.component.tsx
```diff
@@ -99,127 +208,175 @@ const FalseColourPanel = (props: FalseColourPanelProps) => {
...
+            onChangeCommitted={(event, newValue) => {
+              if (Array.isArray(newValue)) {
+                const [lower, upper] = newValue;
+                changeLowerLevel(lower);
+                changeUpperLevel(upper);
               }
             }}
             min={0}
             max={255}
             sx={{
```
louise-davies: We can get rid of all this styling if we remove the `track={false}` bit - as MUI already has the indicators of the selected vs non-selected range. This greying out stuff was needed for the separate sliders to indicate that the min slider must be below the max slider and vice versa, but with them both on the same track this is no longer necessary and we can use the in-built MUI styling

louise-davies (changes requested):
A few comments, and some stuff with the playwright tests that needs to wait for Kenneth's work. But it is working nicely :)

src/images/falseColourPanel.component.tsx
louise-davies: The issue is you need to ensure that the `MenuItem`s are direct children of the `Select`, which can be done like so:

```
{mainColourMapNames.map((mapNames, index) => {
        return [
          <MyListSubheader>{mainColourMapTypeNames[index]}</MyListSubheader>,
          mapNames.map((colourMap) => (
            <MenuItem key={colourMap} value={colourMap}>
              {colourMap}
            </MenuItem>
          )),
        ];
      })}
```

Here, we ensure we're returning arrays, which React parses to make every item a direct descendant.

joshdimanteto: Actually I've got it working using a loop. The Select component was complaining about me using react.Fragment before so I've just implemented a loop without using React.fragment

louise-davies (approved):
LGTM - review comments have been addressed and e2e tests passing now using snapshots generated by docker

PR #254, opened 2023-07-04
DSEGOG-129 display user sessions

joshdimanteto:
## Description
- A user should be able to save a user session and once it's save the session should appear in the drawer (Highlighted). The user should also be able to navigate between session. when a user session is selected it should become highlighted.  

## Testing instructions
- Check if it loads a session correctly
- Check that you can navigate between session without any side effects 
- Check that the list of session updates when a new session is added

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-129


joshdimanteto:
display a save session

joshdimanteto:
- The animation when changing sessions is inconsistent. I'll create another issue so that can be done in another PR

src/mocks/handlers.ts
```diff
@@ -36,6 +37,22 @@ export const handlers = [
     const sessionID = '1';
     return res(ctx.status(200), ctx.json(sessionID));
   }),
+  rest.get('/sessions/list', async (req, res, ctx) => {
+    // const sessionData = sessionsJson.find((session) => session._id === '1');
+    // console.log(sessionData?.session_data);
```
louise-davies: Remove commented out code

src/search/searchBar.component.tsx
```diff
@@ -265,24 +267,32 @@ const SearchBar = (props: SearchBarProps): React.ReactElement => {
...
+    if (sessionId) {
+      setSearchParameterExperiment(null);
+      setSearchParameterFromDate(null);
+      setSearchParameterToDate(null);
+      setSearchParameterShotnumMin(undefined);
+      setSearchParameterShotnumMax(undefined);
+      setMaxShots(MAX_SHOTS_VALUES[0]);
     }
-  }, [dateRange, experimentID, maxShotsParam, shotnumRange]);
+  }, [sessionId]);
```
louise-davies: This code causes a bug where if a user navigates to the `Plots` tab and navigates back - it clears all the search fields. We probably need to not run the hook on mount to avoid this - so probably use a ref to keep track of the first render and not run this code on first render.

src/state/slices/plotSlice.tsx
```diff
@@ -72,7 +72,11 @@ export const plotSlice = createSlice({
     },
     // Use the PayloadAction type to declare the contents of `action.payload`
     closePlot: (state, action: PayloadAction<string>) => {
-      state[action.payload].open = false;
+      // state[action.payload].open = false;
```
louise-davies: Remove commented out code.

louise-davies (changes requested):
A few things need addressing: a few minor comments and a bug I noticed when switching to the Plots tab. But otherwise, the functionality seems to be working well :)

louise-davies (approved):
LGTM

Thought that I had that doesn't necessarily need to be acted on: another way to use the `searchParamsUpdated` function is to rather than pass it to the underlying search components, instead call it in the "search param update functions" e.g. `changeSearchParameterFromDate`. Probably makes the components a little simpler as the prop isn't getting passed through, but `searchBar` gets more complex as we need to create change functions for each parameter where we call `searchParamsUpdated` as well as the `setXParam` functions.

PR #271, opened 2023-07-11
DSEGOG-129 edit and delete a user session

joshdimanteto:
## Description
A user should be able to edit a user session by clicking the pencil icon that is aligned with the session name. This will open a dialog where the user can change the name and the summary. 

A user should be able to delete a user session by clicking the bin icon that is aligned with the session name. This will open a dialog where the user can delete the session. 

## Testing instructions

- [x] successfully deletes a user session
- [x] successfully edits a user session
- [x] error handling 

- [ ] Review code
- [x] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-129


src/app.types.tsx
```diff
@@ -235,6 +235,7 @@ export interface Session {
   summary: string;
   auto_saved: boolean;
   session_data: ImportSessionType;
+  _id?: string;
```
louise-davies: Is `_id` really optional? I assume it's always returned - perhaps we should make it non-optional to ensure we're testing right & if we need to use an `_id` we don't have to null check

src/session/sessionDrawer.component.tsx
```diff
@@ -71,10 +81,24 @@ const SessionListElement = (
...
           </IconButton>
-          <IconButton size="small">
+          <IconButton
+            size="small"
+            onClick={(event) => {
+              event.stopPropagation();
+              openSessionDelete(session);
+            }}
+            data-testid="delete-session-button"
+          >
```
louise-davies: These buttons aren't accessible. Instead of using `data-testid` in tests, use `getByRole("button", {name: "button label")"`. This will then highlight that these buttons aren't accessible since they have no `aria-label`s to provide `button label`

src/session/sessionDrawer.component.tsx
```diff
@@ -23,20 +23,30 @@ const StyledDrawer = styled(Drawer)(({ theme }) => ({
...
+  const {
+    openSessionDelete,
+    openSessionEdit,
+    selected,
+    handleImport,
+    ...session
+  } = props;
 
   return (
     <Box
```
louise-davies: Can we refactor this component to actually use a `List`, and to use the correct `ListItem` components like `ListItemButton`, `ListItemText`. It means we have to do less custom styling (e.g. ListItemButton already has a selected prop) and ensures we're using the correct semantic HTML/accessibility stuff. I have an example of the refactor that I can commit if you'd like :)

louise-davies (changes requested):
Works fine, just a few comments on the code

src/session/sessionDrawer.component.tsx
joshdimanteto: Yeah can you commit the example please :)

src/app.types.tsx
joshdimanteto: The _id is not needed at all. The Session interface is used for the post request which initially does not have an ID anyways

louise-davies (approved):
LGTM

PR #278, opened 2023-07-24
DSEGOG-132 save and save as

joshdimanteto:
## Description
A user should be able to save a user session when they are within a session. A user should also be able to save their current session under a different name and summary. The user session should auto save every 5 minutes and the timestamp should be updated. A new session should be create with "autosaved" appended to the end of the name and then this session will be autosaved whilst in the current session .The session list is sorted so that autosaved session are first and then the most recent.     
 
## Testing instructions

- [ ] Check a user can save their current session and the session data is saved ( navigate to another session and back to check if updates are still there)
- [ ] Check a user can save their current session with a different name and summary 
- [ ] Check that the timestamps update and are in the correct format 
- [ ] Check that the auto save functionality works as expected 
- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-132


src/api/sessions.tsx
```diff
@@ -44,9 +44,11 @@ const editSession = (
   session: SessionResponse
 ): Promise<string> => {
   const queryParams = new URLSearchParams();
+
   queryParams.append('name', session.name);
   queryParams.append('summary', session.summary);
   queryParams.append('auto_saved', session.auto_saved.toString());
+  console.log(session);
```
louise-davies: Remove logging

src/session/sessionSaveButtons.component.tsx
```diff
@@ -1,8 +1,116 @@
...
+
+  const formatDate = (inputDate: string) => {
+    const date = parseISO(inputDate);
+    const formattedDate = format(date, 'dd MMM yyyy HH:mm');
+    return formattedDate;
+  };
+
+  if (selectedSessionTimestamp.timestamp) {
+    timestamp = formatDate(selectedSessionTimestamp.timestamp);
+  }
```
louise-davies: Minor comment, but we could define timestamp all on one line like so:

`const timestamp = selectedSessionTimestamp.timestamp ? formatDate(selectedSessionTimestamp.timestamp) : undefined`

Or, since we do a conditional later on on `timestamp !== undefined`, we could just inline the `formatDate` into the render since we don't use it elsewhere, so the render can be `{selectedSessionTimestamp.timestamp !== undefined ? formatDate(selectedSessionTimestamp.timestamp) : ''}`

src/session/sessionSaveButtons.component.tsx
```diff
@@ -1,8 +1,116 @@
...
+        clearInterval(autoSaveTimer);
+      }
+    };
+  }, [
+    editSession,
+    handleSaveSession,
+    refetchSessionsList,
+    selectedSessionData,
+    sessionId,
+    state,
```
louise-davies: We can't have `state` in the useEffect dependency array - as otherwise the auto-save will likely never trigger (as when a user changes the state, the auto-save timer gets reset. Remove this from dependency array

src/session/sessionSaveButtons.component.tsx
```diff
@@ -12,10 +120,29 @@ const SessionsSaveButtons = () => {
...
+          <span style={{ fontWeight: 'bold' }}>
+            {timestamp !== undefined ? timestamp : ''}
+          </span>
+        </Typography>
+        <Button onClick={handleSaveSession} variant="outlined">
+          Save
+        </Button>
+        <Button onClick={onSaveAsSessionClick} variant="outlined">
+          Save as
+        </Button>
```
louise-davies: Can we have some padding between these two buttons? Currently they're touching

louise-davies (changes requested):
A few comments, plus the auto-save isn't working as described in the [issue ](https://stfc.atlassian.net/browse/DSEGOG-132) (aka auto-saves shouldn't override manual saves, display "auto-saved" sessions at top of session list). However, this could be a separate PR or go into this one, I don't mind.

src/session/sessionSaveButtons.component.tsx
```diff
@@ -1,8 +1,124 @@
...
+
+    return () => {
+      if (autoSaveTimer) {
+        clearInterval(autoSaveTimer);
+      }
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    editSession,
+    handleSaveSession,
```
louise-davies: We also need to delete this from the useEffect, since `handleSaveSession` also changes when `state` changes, which means any change the user makes resets the timer

louise-davies (changes requested):
A few issues I've noticed:

- If the session name overflows, so do the edit and delete buttons. Probably need to apply `whiteSpace: nowrap` to the parent of the buttons
- If you try to delete a session, then the auto-save will now try and auto-save the session you just tried to delete. I believe this is because in `sessionSaveButtons`, we're watching for changes to `selectedSessionId` and not `loadedSessionId`
- If you delete the newly created auto-saved session for the session you have loaded, subsequent attempts to auto-save fail. Perhaps this shows we need to (try to) delete and recreate the auto-saved sessions each time, as there's no guarantee that the user doesn't manually delete the auto-save.

As an aside, I don't know whether we need to auto-save the "blank" session that loads when the user hasn't selected a session - we can ask CLF about this and address it in a later PR though if needed

joshdimanteto:
> A few issues I've noticed:
> 
>     * If the session name overflows, so do the edit and delete buttons. Probably need to apply `whiteSpace: nowrap` to the parent of the buttons
> 
>     * If you try to delete a session, then the auto-save will now try and auto-save the session you just tried to delete. I believe this is because in `sessionSaveButtons`, we're watching for changes to `selectedSessionId` and not `loadedSessionId`
> 
>     * If you delete the newly created auto-saved session for the session you have loaded, subsequent attempts to auto-save fail. Perhaps this shows we need to (try to) delete and recreate the auto-saved sessions each time, as there's no guarantee that the user doesn't manually delete the auto-save.
> 
> 
> As an aside, I don't know whether we need to auto-save the "blank" session that loads when the user hasn't selected a session - we can ask CLF about this and address it in a later PR though if needed



> * If you delete the newly created auto-saved session for the session you have loaded, subsequent attempts to auto-save fail. Perhaps this shows we need to (try to) delete and recreate the auto-saved sessions each time, as there's no guarantee that the user doesn't manually delete the auto-save.

The last bullet point maybe occurring because I'm using `selectedSessionData` instead of `loadedSessionData` in the saveSessionButtons

louise-davies (approved):
LGTM - seems to be working as expected and all my comments have been addressed. We still need to see whether we have to autosave the "empty"/"blank" session but that can be a separate PR/issue

PR #282, opened 2023-09-06
DSEGOG-16 upgrade to react table 8

louise-davies:
## Description
Updates our react-table dependency to v8 - which was a major refactor. All functionality seems to be working fine.

Also, added projection params to the records endpoints that needed them (the table and the plot)

Also, fixed some warnings/errors.

Also, I tried exploring using native pagination stuff from react table (DSEGOG-34), but I don't think it's worth it as the updater function needs to return the new value (which, since we're using redux the dispatch functions don't give us) and so we would only be passing the state in, which is no different than us passing the state manually to the pagination component.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-16


joshdimanteto (approved):
LGTM

PR #297, opened 2023-11-10
DSEGOG-143-hide-empty-rows

kaperoo:
## Description
Rows with no valid data for all displayed columns are hidden from display.
Added the '$or' condition to the search query parameters to check for the existence of data in displayed channels.

## Testing instructions
Updated unit tests to include the '$or' part of the condition in search query parameters comparison.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to DSEGOG-143


src/api/records.tsx
```diff
@@ -68,8 +68,29 @@ const fetchRecords = async (
 
   searchObj.push(...filtersObj);
 
-  if (searchObj.length > 0) {
-    queryParams.append('conditions', JSON.stringify({ $and: searchObj }));
+  const conditions: { [x: string]: { $exists: boolean } }[] = [];
```
louise-davies: Can we rename this to something like `existsConditions` to be more explicit what it is?

louise-davies (changes requested):
I think we also need to apply this change to the `useRecordsCount` query - as otherwise the pagination component which tells you how many items there are in the table (e.g. if unlimited is selected or 1000 is selected and count < 1000) would be wrong - this leads to there being "empty" pages in the table

src/api/records.tsx
```diff
@@ -585,15 +625,22 @@ export const useIncomingRecordCount = (
...
+      { searchParams: SearchParams; filters: string[]; projection: string[] },
+    ]
   >(
     [
       'incomingRecordCount',
-      { searchParams: finalisedSearchParams, filters: finalisedFilters },
+      {
+        searchParams: finalisedSearchParams,
+        filters: finalisedFilters,
+        projection,
```
louise-davies: I think we need to remove the projection from the incoming record count query, as otherwise we'd need to check the user isn't querying for too many rows whenever they change the data channels they're viewing.

A slight optimisation we can do is to keep the `projection` key here in the query key, but set it to `projection: [timeChannelName]` - this way we  can still optimise the `recordCount` query in the "blank table" use case, but otherwise we don't use the projection elsewhere in `incomingRecordCount`

louise-davies (approved):

PR #301, opened 2023-12-06
fix table pagination bug

kaperoo:
## Description
Fix for DSEGOG-281. 
Reset pagination to 0 whenever requested page data is changed, but not the page number itself.
It will also reset to the first page if the page is out of _'max shot'_ range.

## Testing instructions
Change channels/shots/filters to see if pagination is reset to 0.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to DSEGOG-281


louise-davies (approved):
LGTM! Page seems to be resetting to 0 in all situations it should do (filtering, new searches, change in rows per page, change in results count etc.)

PR #302, opened 2023-12-13
row selection + some styling

kaperoo:
## Description
Add row selection to the table.

Since we use server-side pagination, it is not possible to select all rows with TanStack (known [issue](https://github.com/TanStack/table/issues/4781)).
For now the checkbox in the header of the table toggles selection for all rows on the current page. 

Also, added a bit of styling to the table cells so that everything is aligned:
image

Temporarily (?) switched off the background colour of the columns so that the highlighting of selected rows is visible.

## Testing instructions
Selecting rows on one page shouldn't select (or unselect) any rows on any other page.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to DSEGOG-120


src/table/table.component.tsx
```diff
@@ -154,6 +163,22 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
                           ref={provided.innerRef}
                           sx={{ display: 'flex', flexDirection: 'row' }}
                         >
+                          <MuiTableCell
+                            padding="checkbox"
+                            sx={{ width: '48px' }}
```
louise-davies: Removing these manually set widths doesn't affect the styling at all as far as I can see, so we should probably remove them

src/table/table.component.tsx
```diff
@@ -215,16 +240,31 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
                 <MuiTableRow
-                  key={row.id}
+                  key={row.original._id}
                   sx={{ display: 'flex', flexDirection: 'row' }}
+                  selected={row.getIsSelected()}
+                  aria-checked={row.getIsSelected()}
                 >
+                  <MuiTableCell
+                    padding="checkbox"
+                    sx={{ width: '48px' }}
```
louise-davies: See above

src/table/table.component.tsx
```diff
@@ -215,16 +240,31 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
+                    scope="row"
+                  >
+                    <Checkbox
```
louise-davies: These checkboxes look a little big/have too much padding imo - especially in a text only table. Can we experiment with size small checkboxes and reducing their padding (which reduced the size of the ripple effect)

Also, they aren't vertically aligned when an image or trace channel is present

kaperoo:
Changed checkbox size to _small_ and reduced their padding:
image

Also, checkboxes are now vertically aligned:
image

src/table/table.component.tsx
```diff
@@ -239,16 +264,36 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
+                    padding="none"
+                    sx={{
+                      alignItems: 'center',
+                      display: 'flex',
+                    }}
+                  >
+                    <Checkbox
+                      size="small"
+                      inputProps={{ 'aria-label': 'select row' }}
+                      sx={{ paddingTop: '5px', paddingBottom: '5px' }}
```
louise-davies: Minor nitpick, but this messes with the ripple shape (see below)

image

I believe if we set padding on all sides to 5, and then add back in the extra space on the left and right with margin 4px it fixes it e.g.
`sx={{ p: "5px", m: "0 4px" }}`

image

kaperoo: Yeah, I see it now. I'll also add padding + margins to the _"select all"_ checkbox so that the ripple sizes are the same.

louise-davies (approved):
Kacper has tested my code and both the sticky columns & the row selection seem to be working well with all the other table features now

PR #315, opened 2023-12-15
Improve cypress stability

louise-davies:
## Description
I noticed that some cypress tests were flaky due to failing in the beforeEach hook because I guess waiting for progress bar to appear then disappear could cause issues if the page loads too fast for cypress to see the progress bar? Now we just use another assert to check the page has loaded and check the progress bar is not visible, which I think should help stabilise these tests

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


kaperoo (approved):
Looks good to me 👍

PR #345, opened 2024-01-26
Fix traces

louise-davies:
## Description
Fix the waveform API fetch function as we now get an array of numbers rather than a string from the API so no need to parse anymore

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


kaperoo (approved):
Looks good to me 👍

PR #364, opened 2024-03-01
DSEGOG-119 DSEGOG-120 Data export

kaperoo:
## Description
This PR build on #302 and adds the export dialogue and the actual export functionality. 
It also (temporarily) fixes DSEGOG-141 by sorting by timestamp in ascending order if no other sort exists.

New _Export_ button next to session save buttons:
image

The export dialogue with all the export options:
image
(Currently exporting _Selected Rows_ with no rows selected just throws an error in the console).

Note that the record selection is saved in the session.

## Testing instructions
Run the updated unit and e2e tests.
Also, added new testcases: `export.test.tsx` and `export.cy.ts`.

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
connect to DSEGOG-119


cypress/e2e/export.cy.ts
```diff
@@ -0,0 +1,88 @@
...
+    cy.findByRole('button', { name: 'Export' }).click();
+    cy.findByRole('dialog').should('exist');
+  });
+
+  it('closes the export dialogue', () => {
+    cy.findByRole('button', { name: 'Export' }).click();
+    cy.findByRole('dialog').should('exist');
+    cy.findByRole('button', { name: 'Cancel' }).click();
+    cy.findByRole('dialog').should('not.exist');
+  });
```
louise-davies: This can be one "can open and close the dialog" test since the first test case is covered in the latter anyway

src/api/export.tsx
```diff
@@ -0,0 +1,183 @@
...
+        dataToExport: DataToExport;
+      };
+      const startIndex =
+        exportType === 'Visible Rows' ? page * resultsPerPage : 0;
+      const stopIndex =
+        exportType === 'Visible Rows'
+          ? startIndex + resultsPerPage
+          : exportType === 'All Rows'
+            ? maxShots
+            : 0;
```
louise-davies, line 171: This logic doesn't work when maxShots is Unlimited i.e. the value of `maxShots` is `Infinity` - the backend doesn't like `Infinity` as a value. We should probably not set `offsetParams` at all in that case

louise-davies: We should also probably have a test case to cover Unlimited shots since handling with `Infinity` is weirder than the difference between 50 and 1000

src/export/exportButton.component.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+    });
+  });
+
+  it('opens export dialogue on button click', () => {
+    renderComponentWithProviders(<ExportButton />);
+    const exportButton = screen.getByText('Export');
+    fireEvent.click(exportButton);
+    const exportDialogue = screen.getByRole('dialog');
+    expect(exportDialogue).toBeInTheDocument();
+  });
```
louise-davies: This test seems to be covered by the previous test case

src/export/exportButton.component.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { screen, fireEvent, waitFor } from '@testing-library/react';
+import { renderComponentWithProviders } from '../setupTests';
+import ExportButton from './exportButton.component';
+import React from 'react';
+
+describe('ExportButton', () => {
+  it('opens and closes the dialog on button clicks', async () => {
+    renderComponentWithProviders(<ExportButton />);
+
+    fireEvent.click(screen.getByText('Export'));
```
louise-davies: Is there a reason `fireEvent` is needed here over `userEvent`?

src/export/exportDialogue.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+export interface ExportDialogueProps {
+  open: boolean;
+  onClose: () => void;
+}
+
+const ExportDialogue = (props: ExportDialogueProps) => {
+  const { open, onClose } = props;
+
+  const mutate = useExportData().mutate;
```
louise-davies: Style-wise, I think `const { mutate } = useExportData();` looks better, plus is more extensible if we ever need other variables from `useExportData`

src/export/exportDialogue.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const [selectedExportContent, setSelectedExportContent] = React.useState({
+    Scalars: true,
+    Images: false,
+    'Waveform CSVs': false,
+    'Waveform Images': false,
+  });
+
+  const handleClose = React.useCallback(() => {
+    onClose();
+  }, [onClose]);
```
louise-davies: This wrapper function isn't needed as `handleClose` as the exact same call signature as `onClose` - just pass `onClose` directly where you pass `handleClose`

src/views/recordTable.component.test.tsx
```diff
@@ -99,6 +99,61 @@ describe('Record Table', () => {
...
+      expect(store.getState().selection.selectedRows).toEqual([]);
+    });
+
+    it('can select single and multiple rows', async () => {
+      const user = userEvent.setup();
+      server.use(
+        rest.get('/records', (req, res, ctx) => {
+          return res(ctx.status(200), ctx.json(recordsJson.slice(0, 3)));
+        })
+      );
```
louise-davies: Why is this necessary? I get in the previous test we're reducing the number of results to make asserting all rows selected easier, but since in this test we're just indexing individual rows I would guess the test works fine without this

louise-davies (changes requested):
Overall looking good - just found 1 bug to do with Unlimited shots and various code style comments.

Another comment is that if the export request takes a long time - there's no visual to the user that the request is processing. Perhaps we can use a [LoadingButton](https://mui.com/material-ui/react-button/#loading-button)? Would also need to disable clicking away from the export dialogue when fetching as well... Alternatively, eCat has a "generating data" overlay which we could also do with a [Backdrop](https://mui.com/material-ui/react-backdrop/)

image

kaperoo:
Addressed comments, added a _"generating export data"_ backdrop and fixed the default timestamp sorting issue.

|Backdrop in light mode:|Backdrop in dark mode:|
|:-------------------------:|:-------------------------:|
|light|dark|

louise-davies (approved):
Seems to be working well - all rows export is working with Unlimited shots mode and the loading dialogue looks good :)

PR #375, opened 2024-06-19
DSEGOG-296 add e2e test suite against real API

louise-davies:
## Description
Adds an additional playwright test suite that runs against a real API instance to test integration between the front and backend. Note we only test things that test how we communicate with the backend is correct - any extra testing should be done in the mocked e2e tests only.

Most of the file changes are from moving the existing playwright tests into a subfolder

There's one test that doesn't pass when ran against the dev server instead of the CI data - so that test is skipped when not in a CI environment.

Matt said he's willing to maintain the "set up the backend" part of the new workflow

TODO: once this is merged in, re-instate the branch protection rules for the new jobs

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] See if you can run the tests locally against the dev server

## Agile board tracking
Closes DSEGOG-296


joelvdavies:
We found something I thought would be worth raising while trying to run the standard playwright tests initially. When running the e2e tests on a rocky 8 VM we had an permission error on `mkdir` while it tried to create the `playwright-report` folder. This was ultimately resolved it by running `chmod -R a+rw .` to change the permissions of the whole folder. It didn't happen on WSL and it seems the rocky 8 machine used a different user for docker.

While debugging this Viktor pointed out this may have been easier to debug if the tests had their own Dockerfile rather than being on a single line and also only copying in only the required files instead of mounting everything.

e2e/real/sessions.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  ).not.toBeVisible();
+
+  await expect(
+    page.getByRole('button', {
+      name: 'renamed e2e testing session',
+      exact: true,
+    })
+  ).not.toBeVisible();
+
+  // if we get to the end of the test, no need to our after each to delete the session
```
joelvdavies: This comment doesn't make sense.

joshdimanteto (dismissed):
LGTM

e2e/real/sessions.spec.ts
louise-davies: Good catch, took me a second to think of what I was meaning to say with that comment: `if we get to the end of the test, no need to run our after each code to delete the session (since we just deleted it as part of the test), so reset session id`. I'll update it in a sec

louise-davies:
> We found something I thought would be worth raising while trying to run the standard playwright tests initially. When running the e2e tests on a rocky 8 VM we had an permission error on `mkdir` while it tried to create the `playwright-report` folder. This was ultimately resolved it by running `chmod -R a+rw .` to change the permissions of the whole folder. It didn't happen on WSL and it seems the rocky 8 machine used a different user for docker.
> 
> While debugging this Viktor pointed out this may have been easier to debug if the tests had their own Dockerfile rather than being on a single line and also only copying in only the required files instead of mounting everything.

@joelvdavies @joshuadkitenge I think I figured out a fix - if I set it so the container runs as the current user then it works on Linux (both Rocky 8 cloud VM and WSL) (we still use pwuser on Windows but that doesn't have permissions issues so it's fine). As I said, I want to avoid creating a custom docker container as all we want to do is run playwright in the container and ensure it has read & write access to our test code so it can read tests & write back snapshots etc. We would likely still have permissions problems if we had a separate Dockerfile, plus then you'd have to remember to rebuild the Dockerfile each time, which we did previously have a separate Dockerfile and it made running the tests quite annoying.

I think the reason why it worked on WSL is that by default the first user created in WSL is uid 1000, which is the linux default "first user" and so matches the user created in the PW docker container.

The new commands now work on Windows, WSL and a cloud Rocky 8 machine (or at least those are the envs I have tested the new commands against).

joelvdavies:
> @joelvdavies @joshuadkitenge I think I figured out a fix - if I set it so the container runs as the current user then it works on Linux (both Rocky 8 cloud VM and WSL) (we still use pwuser on Windows but that doesn't have permissions issues so it's fine). As I said, I want to avoid creating a custom docker container as all we want to do is run playwright in the container and ensure it has read & write access to our test code so it can read tests & write back snapshots etc. We would likely still have permissions problems if we had a separate Dockerfile, plus then you'd have to remember to rebuild the Dockerfile each time, which we did previously have a separate Dockerfile and it made running the tests quite annoying.
> 
> I think the reason why it worked on WSL is that by default the first user created in WSL is uid 1000, which is the linux default "first user" and so matches the user created in the PW docker container.
> 
> The new commands now work on Windows, WSL and a cloud Rocky 8 machine (or at least those are the envs I have tested the new commands against).

Yeah, fair enough, I think using `docker build` in the command before the run could also work but it would take longer to run if there were any image changes and I agree it would also cause a permission issue with the reports as well. 

I tested it on my Rocky 8 VM and it worked without any permissions issues 🥳 . Thanks for looking into it. Some of the tests still seem a bit flaky locally but it seems that's just me, the mocked ones are the same and they all pass fine on the CI.

joelvdavies (approved):

PR #376, opened 2024-07-10
DSEGOG-109 DSEGOG-110 DSEGOG-112 Functions 

joshdimanteto:
## Description

The ability to create functions and display functions data in the tables  
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking

closes  DSEGOG-109 , DSEGOG-112


joshdimanteto:
The E2E real test are not going to work until https://github.com/ral-facilities/operationsgateway-api/pull/86 is merged

package.json
```diff
@@ -22,6 +22,7 @@
     "@types/node": "20.14.10",
     "@types/react": "18.2.48",
     "@types/react-dom": "18.2.18",
+    "@types/uuid": "^10.0.0",
```
joelvdavies: Is this required if you are using crypto?

joelvdavies:
I think these should work
image
image

cypress/e2e/functions.cy.ts
```diff
@@ -0,0 +1,102 @@
+describe('Sessions', () => {
```
louise-davies: Copy paste error ;)

src/api/channels.test.tsx
```diff
@@ -75,6 +80,160 @@ describe('channels api functions', () => {
       });
     });
 
+    it('uses a select function to construct an array of columns from given channel and functions metadata (scalar)', async () => {
```
louise-davies: Minor comment that I don't see why we separated the scalar function channel from the image & waveform function channels since we're just checking the conversion function

src/api/functions.tsx
```diff
@@ -0,0 +1,100 @@
...
+        .replace(/\s*\(\s*/g, '(')
+        // Remove any spaces around the close parenthesis ')'
+        .replace(/\s*\)\s*/g, ')')
+        // Add a space between operators (*, +, -, /) and the following open parenthesis '('
+        .replace(/([*+\-/])\(/g, '$1 (')
+        // Add a space between the close parenthesis ')' and the following operators (*, +, -, /)
+        .replace(/\)([*+\-/])/g, ') $1')
+        // Remove any spaces around the double asterisk '**' (exponentiation operator)
+        .replace(/\s*\*\*\s*/g, '**'),
+    })),
```
louise-davies, line 41: These comments are very descriptive - very good when dealing with regex!

src/api/functions.tsx
```diff
@@ -0,0 +1,100 @@
...
+      headers: {
+        Authorization: `Bearer ${readSciGatewayToken()}`,
+      },
+    })
+    .then((response) => {
+      return response.data;
+    });
+};
+
+export const useGetFunctionsTokens = (): UseQueryResult<
```
louise-davies: Minor comment, `useGetFunctionTokens` and `usePostValidateFunctions` can probably drop the `Get` and `Post` as that's just referring to the HTTP keyword right? `useFunctionTokens` and `useValidateFunctions` are probably more in line with our current naming for react query hooks

`usePostValidateFunctions` is extra confusing as it kind of sounds like "post-validate" aka after validating

src/api/records.tsx
```diff
@@ -534,20 +583,29 @@ export const useThumbnails = (
...
   const { apiUrl } = useAppSelector(selectUrls);
-  const { searchParams, filters } = useAppSelector(selectQueryParams);
+  const { searchParams, filters, functions } =
+    useAppSelector(selectQueryParams);
   const queryClient = useQueryClient();
   const projection = useAppSelector(selectSelectedIdsIgnoreOrder);
 
   return useQuery({
-    queryKey: ['recordCount', { searchParams, filters, projection }],
+    queryKey: ['recordCount', { searchParams, filters, projection, functions }],
```
louise-davies, line 589: A question - the only reason we're adding functions here is because we need to account for function channels in the exists conditions correct? as otherwise functions don't affect the record count. Just wanted to make sure I understand :)

src/app.types.tsx
```diff
@@ -272,3 +272,42 @@ export interface SessionListItem {
   timestamp: string;
   _id: string;
 }
+
+export interface FunctionTag {
+  type: 'channel' | 'functionToken' | 'function' | 'number' | 'string';
```
louise-davies: Is `string` a valid function tag? I thought plain strings weren't allowed in functions

src/functions/functionsDialog.component.tsx
```diff
@@ -0,0 +1,363 @@
...
+};
+
+const FunctionsDialog = (props: FunctionsDialogProps) => {
+  const { open, onClose, flashingFunctionValue } = props;
+
+  const dispatch = useAppDispatch();
+  const appliedFunctions = useAppSelector(selectAppliedFunctions);
+  const appliedSelectedIds = useAppSelector(selectSelectedIds);
+  // we need searchParams so we can check for past queries before showing the warning message
+  // const searchParams = useAppSelector(selectSearchParams);
```
louise-davies: Remove commented out code and outdated comment

src/app.types.tsx
```diff
@@ -272,3 +272,42 @@ export interface SessionListItem {
...
+  value: string;
+  label: string;
+}
+
+export interface FunctionToken {
```
louise-davies: I wonder in naming wise, to keep consistency with filters this should be `FunctionOperator`, and thus `FunctionTag` can be `FunctionToken` (which is more equivalent to the `Token` type used in filters).

src/functions/functionsDialog.component.tsx
```diff
@@ -0,0 +1,363 @@
...
+      postValidateFunctions,
+      functions,
+      handleError,
+      dispatch,
+      appliedSelectedIds,
+      onClose,
+    ]
+  );
+
+  const functionsToken: FunctionTag[] = functions
```
louise-davies: Confusingly named - maybe `tokenisedFunctions`? Made even worse with the token/tag crossover here

louise-davies (commented):
Overall looks very good, just some minor comments. Functionality is all working well - the only thing I noticed is we can't filter functions. I'm not sure if this is a backend limitation or we just didn't include it here. Note: eCat doesn't support filtering on functions, but it does say: `Function names are not currently supported in filters`.

src/api/records.tsx
joshdimanteto: Yes, that is the reason

joelvdavies (dismissed):
LGTM Reviewed the changes from Louise's and my last review. I think develop still needs to be merged in as I found a reference to jest fixed by my last PR, either before this is merged, or just in the next PRs on top of this one. Also side note https://github.com/ral-facilities/operationsgateway/pull/430 should be tested with this change.

joelvdavies (approved):

PR #421, opened 2024-07-26
DSEGOG-341 Migrate from CRA to Vite

joelvdavies:
## Description
Migrates from CRA to Vite. Following on from https://github.com/ral-facilities/inventory-management-system/pull/348 and https://github.com/ral-facilities/scigateway/pull/1380.

### Notes
- Copied new `tsconfig` from IMS (The config here was copied to IMS in the first place)
- Instances of process.env.REACT_ replaced with import.meta.env.VITE_
  - Renamed `VITE_APP_OPERATIONSGATEWAY_BUILD_DIRECTORY` to `VITE_APP_OPERATIONS_GATEWAY_BUILD_DIRECTORY`
  - Replaced `REACT_APP_E2E_TESTING` with `VITE_APP_INCLUDE_MSW` and `VITE_APP_BUILD_STANDALONE` (Same as IMS so could use the same config and allows more explicit differentiation of e2e tests with and without the api)
- Some handlers in handlers.ts don't match the backend e.g. `/sessions/:id` doesn't have a duplicate check in the backend when it probably should (also on a related note the post's should really use 201's but the backend currently uses 200s.)
- Like for IMS I have renamed `setupTests.tsx` to `testUtils.tsx` and moved MSW setup to a separate new `setupTests.ts` imported in the vitest config - to avoid repeated execution during imports)
- Replaced `jest-canvas-mock` with `vitest-canvas-mock`
- MSW v2 stores the url as a string in responses, so many tests that directly accessed that have been modified to first convert to a URL object before accessing the search params
- Removed `applyDatePickerWorkaround` and `cleanupDatePickerWorkaround` from unit tests - I added this in datagateway for MUI v5 when we enzyme and it was copied over here, but doesn't seem like its needed anymore (and broke here anyway)
- Like IMS, removed `cypress-failed-log` as its 2 years old and I don't believe it does anything anymore or otherwise the import is broken as it cannot use require
- Moved msw to be a full dependency instead of a dev one - This is like IMS as the `yarn tsc` in the build process uses the `tsconfig.build.json`. While the mock handlers could be excluded from this, it will then also break any e2e tests relying on this build - the fix would be to have another tsconfig for this case, and adding another build command that is only used in this case. Given it should already be removed from the bundle at the end, I thought this was simpler.
- `yarn lint` is not the same as on IMS - we include tests in the checks with tsc to resolve further typescript issues, but there are many here that have been ignored and I thought it best to leave to a further eslint v9 PR if addressed as a result
- Was originally going to upgrade to eslint v9 at the same time as I thought the changes to eslint were more complicated, but was simple in practice so leaving to another PR to avoid making more changes here

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-341


package.json
```diff
@@ -73,42 +77,28 @@
...
-    "resetMocks": false,
-    "transformIgnorePatterns": [
-      "node_modules/(?!copy-anything|axios|is-what)"
-    ]
-  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not op_mini all"
+  ],
```
louise-davies: Our browserlist could technically be just chrome, firefox & safari as that's what our requirements say: https://stfc.atlassian.net/browse/DSEGOG-84?atlOrigin=eyJpIjoiMDMzODBkZjgxNjVhNDJhMzhlOWFlMGFiZmU0YzM1NmEiLCJwIjoiaiJ9 but I don't think I really object to keeping the "standard" browserslist - or could even just use "defaults" from browserslist

src/images/imageWindow.component.test.tsx
```diff
@@ -1,28 +1,29 @@
...
-import { server } from '../mocks/server';
-import { TraceOrImageWindow } from '../state/slices/windowSlice';
 import { DEFAULT_WINDOW_VARS } from '../app.types';
+import { TraceOrImageWindow } from '../state/slices/windowSlice';
+import { renderComponentWithProviders } from '../testUtils';
+import ImageWindow from './imageWindow.component';
 
-jest.mock('../windows/windowPortal.component', () => {
+vi.mock('../windows/windowPortal.component', async () => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
```
louise-davies: These eslint disable lines can now be removed

src/main.tsx
```diff
@@ -184,51 +183,62 @@ export const fetchSettings = (): Promise<OperationsGatewaySettings | void> => {
...
-      // need to use require instead of import as import breaks when loaded in SG
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      const { worker } = require('./mocks/browser');
-      return worker.start();
-    } else return Promise.resolve();
+    if (
+      import.meta.env.VITE_APP_INCLUDE_MSW === 'true' ||
+      settingsResult?.apiUrl === ''
+    ) {
+      // Need to use require instead of import as import breaks when loaded in SG
```
louise-davies: This comment needs removing as it's now contradictory

src/plotting/plot.component.tsx
```diff
@@ -1,16 +1,15 @@
...
-  YAxisScale,
   SelectedPlotChannel,
+  XAxisScale,
+  YAxisScale,
 } from '../app.types';
 // only import types as we don't actually run any chart.js code in React
-import type { ChartOptions, ChartDataset } from 'chart.js';
+import type { ChartDataset, ChartOptions } from 'chart.js';
 // we import this even though we don't use it so we can get typescript info added to ChartOptions
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
louise-davies: Remove redundant lines

src/plotting/plotSettings/plotSettingsController.component.test.tsx
```diff
@@ -1,33 +1,32 @@
...
+  const changeLeftYAxisMinimum = vi.fn();
+  const changeLeftYAxisMaximum = vi.fn();
+  const changeRightYAxisMinimum = vi.fn();
+  const changeRightYAxisMaximum = vi.fn();
+  const changePlotTitle = vi.fn();
+
+  const createView = async () => {
     // need to import like this in order for the doMock's to work
     const PlotSettingsController =
       // eslint-disable-next-line @typescript-eslint/no-var-requires
```
louise-davies: Remove as now doesn't apply

src/plotting/plotSettings/yAxisTab.component.tsx
```diff
@@ -472,7 +472,7 @@ const YAxisTab = (props: YAxisTabProps) => {
                   // for some reason, styling these buttons in a row causes webkit
                   // headless playwright e2e tests on linux to fail - so disable this styling in e2e builds
                   /* istanbul ignore next */
-                  process.env.REACT_APP_E2E_TESTING === 'true'
+                  import.meta.env.VITE_APP_BUILD_STANDALONE === 'true'
```
louise-davies: Might be worth checking if this still applies with browser updates etc.

i.e. remove work around code and check playwright mock tests run correctly on webkit

src/testUtils.tsx
```diff
@@ -5,96 +5,100 @@ import React from 'react';
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
```
louise-davies: This was imported in `setupTests.tsx` right? So probably doesn't need to also be imported here

src/testUtils.tsx
```diff
@@ -5,96 +5,100 @@ import React from 'react';
...
-// Clean up after the tests are finished.
-afterAll(() => server.close());
-
 /**
  * Waits for msw request -
  * @param method string representing the HTTP method
  * @param url string representing the URL match for the route
  * @returns a promise of the matching request
  *  */
+// TODO: Could this be replaced using an axios spy for the search params instead?
```
louise-davies: Have you investigated this TODO or is it best left here as a future improvement?

src/traces/tracePlot.component.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Waveform } from '../app.types';
 // only import types as we don't actually run any chart.js code in React
-import type { ChartOptions, ChartData } from 'chart.js';
+import type { ChartData, ChartOptions } from 'chart.js';
 // we import this even though we don't use it so we can get typescript info added to ChartOptions
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
louise-davies: Remove unneeded lines

vite.config.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  return {
+    ...config,
+    test: {
+      globals: true,
+      environment: 'jsdom',
+      globalSetup: './globalSetup.js',
+      setupFiles: ['src/setupTests.ts'],
+      exclude: [...vitestDefaultExclude, 'e2e/**'],
+      coverage: {
```
louise-davies, line 147: According to https://vitest.dev/guide/coverage.html#coverage-providers - `v8` is now the default reporter so probably need to find/replace any istnbul ignore comments with v8 ignore (https://vitest.dev/guide/coverage.html#ignoring-code)

louise-davies (commented):
The code is looking good aside from some minor comments of mine. Dev mode works great, as does tests. However, when running `yarn preview:dev:build` - I can't load OG in the SG react 18 branch. I'm getting an `Uncaught SyntaxError: import declarations may only appear at top level of a module` error, and it's pointing to a React import. Might be me doing something wrong but thought I'd raise it!

joelvdavies:
> The code is looking good aside from some minor comments of mine. Dev mode works great, as does tests. However, when running `yarn preview:dev:build` - I can't load OG in the SG react 18 branch. I'm getting an `Uncaught SyntaxError: import declarations may only appear at top level of a module` error, and it's pointing to a React import. Might be me doing something wrong but thought I'd raise it!

Will address the other comments in the morning. Just spent a while debugging this with Joshua and finally got to the cause. Its the same issue you noticed on IMS, only there it was intermittent - almost every other build it did came out with imports in the main.js that shouldn't have been there breaking the build, only it seems that in the past few months an update has caused it to become a permanent issue not an intermittent one. I now believe as I was not specifying a format in the build, it used a default of `es` and `umd`. But due to the output filename these were both output to the same file. By forcing it to one of them at a time I have found that yarn build does this too if I use `es` instead of `umd`, so I think the IMS issue was caused by them producing both files and depending on speed each would be used only some of the time. Now `yarn build` seems to cause it to use `umd` and `build --watch` uses `es`. I am not clear on why the es modules build is doing this when it builds as a library - as far as I am aware it should be capable of it as I only used the example from the vite documentation. But for now the obvious solution is to use umd for the build for scigateway. So es modules are only used through dev (where the speed is beneficial) and in scigateway where this doesn't cause a problem. Hopefully a future update will fix whatever this is (although it could be my own misunderstanding of the config as well).

src/testUtils.tsx
joelvdavies: Ah no I didn't, could be a future improvement? I wasn't sure if there was a reason we wait like this. I think Joshua and I discussed it at some point but that was a while ago now. Happy to remove it or leave it in for the future.

vite.config.ts
joelvdavies: Nice spot, seems like there was only one ignore.

louise-davies:
@joelvdavies I believe the issue with the es build is we're not expecting an es module when loading the plugin from SciGateway - you need special handling to be able to handle es modules (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#applying_the_module_to_your_html)

If we really wanted es modules, it's probably worth waiting until everything is migrated over to Vite, and then overhauling the SG plugin loading system to be able to handle es modules (either in replacement of handling non-modules, or be able to load either using a feature switch in the plugin config?)

Either way, switching to UMD solves the immediate problem 👍

src/testUtils.tsx
joelvdavies: As discussed will remove as it was done this way for a reason.

src/plotting/plotSettings/yAxisTab.component.tsx
joelvdavies: No longer needed, have removed it.

package.json
joelvdavies: As discussed going to defaults for simplicity.

louise-davies (approved):
With the changes to the build, I'm happy that everything works (i.e. dev, loading production build in SG, e2e tests, unit tests) - at least for me.

joshdimanteto (approved):
LGTM tested dev, loading production build in SG, e2e tests, unit tests and e2e tests

PR #427, opened 2024-08-06
Add checkboxes to functions

joshdimanteto:
## Description
- Add checkbox to functions dialog to determined if they should be visible
- Make the dialog bigger and reduce the size of the name input and
  the help section
- Make the operators in the help pages chip instead of plain text

 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}


src/functions/functionsDialog.component.tsx
```diff
@@ -268,16 +291,54 @@ const FunctionsDialog = (props: FunctionsDialogProps) => {
...
             <Heading mt={1}>Enter function</Heading>
             {functions.map((func, index) => {
               return (
-                <Grid container item key={func.id}>
+                <Grid pl={0} container item key={func.id}>
+                  <Grid item>
+                    <Tooltip
+                      title={
+                        selectedColIds.includes(func.id)
+                          ? 'Removed function column'
```
louise-davies: `Removed` -> `Remove function column` (or maybe even `Hide` to contrast better with `Display`?)

src/functions/functionsDialog.component.tsx
```diff
@@ -340,7 +406,7 @@ const FunctionsDialog = (props: FunctionsDialogProps) => {
             </Grid>
           </Grid>
           <Divider orientation="vertical" flexItem />
-          <Grid item xs>
+          <Grid item sx={{ width: 650 }}>
```
louise-davies: Maybe this needs to be a `maxWidth` or something as otherwise it takes up a lot of room on smaller screens.

louise-davies (dismissed):
If you check/uncheck a function and then close the dialogue, the change isn't applied but if you re-open the dialogue the checkbox doesn't match the visibility state (i.e. it remembered that you clicked the checkbox but didn't apply the change). Either need to apply the change on click or probably more sensibly reset the checkboxes to the redux visibility state on dialogue close (without applying)

joelvdavies (commented):
Just on Louise's point, not sure how small we expect it to get, but before it was quite nice if it got too small as it would go on a new line
image

whereas now its stuck as xs 3, could you use a hybrid like the IMS subsystems so it goes underneath on small screens and takes up a fraction otherwise. I think I used `xs={12} md` there.

joelvdavies (approved):

PR #429, opened 2024-08-14
DSEGOG-342 Misc fixes

louise-davies:
## Description
Fixes a few misc issues either raised in the previous user feedback session or found just before it:
- Sets a default timeframe of 24 hours to ensure we don't query for all of the data on page load!
- Removes anti-aliasing from image zooming
- Saves the position of popups on session save. Previously, window positions of plots were only saved on plot save, and image/traces window positions were only saved if you used the thumbnail selector!
- Improve styling (e.g. searchbar is now more flex-y, means we don't cut off shot numbers in exchange for overflowing when the screen gets to small)
- Fix infinite shots not loading from sessions correctly
- Added logo
- Removed webkit playwright skips as they seem to be fixed now :)
- Added clearable prop to datepickers which adds a clear button to the text input

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-342


cypress.config.js
```diff
@@ -14,6 +14,21 @@ module.exports = defineConfig({
       on('task', { removeDirectory });
       // https://github.com/bahmutov/cypress-failed-log
       require('cypress-failed-log/on')(on);
+
+      on('before:browser:launch', (browser, launchOptions) => {
+        if (browser.family === 'chromium' && browser.name !== 'electron') {
+          // Set pointer type to fine so that date inputs work properly
```
joelvdavies, line 20: Just wondering what the actual issue was that these launch options fix? Perhaps the comment could be updated to explain it?

playwright.config.ts
```diff
@@ -68,6 +68,27 @@ const config: PlaywrightTestConfig = {
...
+          dependencies: ['setup'],
+        },
+        // sessions test needs to run on firefox to test saving position & size of popups
+        {
+          name: 'E2E tests - Firefox',
+          use: {
+            ...devices['Desktop Firefox'],
+            launchOptions: {
+              // need these to ensure Date picker media queries pass
+              // ref: https://mui.com/x/react-date-pickers/getting-started/#testing-caveats
```
joelvdavies: I think this information has been moved to https://mui.com/x/react-date-pickers/base-concepts/#testing-caveats

src/operationsgateway-logo-white.png
joelvdavies, line 1: Very minor point, but this looks kind of blury to me compared to datagateway which was an svg. Also the `Gateway` was previously a different font. Presumably we don't have the original file anymore?

joelvdavies (commented):
Couple very minor comments. Also just wanted to check that the window position saving isn't supposed to work across multiple screens?

louise-davies:
@joelvdavies multi-screen window saving is done at a "best effort" basis. Because we're working with browser APIs, it's hard to truly manipulate things exactly. If you position a popup slightly offscreen, then `window.open` will shift it so it's entire on screen for example. For multi-monitors, dealing with e.g. different screen sizes is I think what messes the code up. I can make it slightly better at positioning things when the x & y co-ords are negative (we then have to take into account the height & width as it's measured from top & left), but it will still get things "wrong" e.g. if one monitor is smaller/bigger than the other, if you change your primary monitor to another monitor etc.

I'll add the better handling of negative co-ords, but otherwise it's not really feasible to cover other usecases...

joelvdavies:
> If you position a popup slightly offscreen, then `window.open` will shift it so it's entire on screen for example. For multi-monitors, dealing with e.g. different screen sizes is I think what messes the code up. I can make it slightly better at positioning things when the x & y co-ords are negative (we then have to take into account the height & width as it's measured from top & left), but it will still get things "wrong" e.g. if one monitor is smaller/bigger than the other, if you change your primary monitor to another monitor etc.

Yeah sure, no worries I thought that was likely the case. This is still better than nothing.

cypress.config.js
louise-davies: It makes the datepickers editible as cypress doesn't allow for mocking media queries - I stole the code from DGW: (https://github.com/ral-facilities/datagateway/blob/develop/packages/datagateway-dataview/cypress.config.ts) but it's similar to the launch options in playwright config. I can add a link to the testing caveats like we do in the PW config

src/operationsgateway-logo-white.png
louise-davies: We do still have the `.ai` file for the non-white version - would need to convert it to SVG then make the white text version. I'll have a go :)

joelvdavies (approved):
LGTM

joshdimanteto (approved):
LGTM

PR #430, opened 2024-08-16
DSEGOG-345 Fix bug where changing session then immediately changing filters sets the search params back to defaults

louise-davies:
## Description
See commit message for details. But I found a bug that was made worse with my bugfixes PR with the change so that the default is a non-null daterange. If you change the filters directly after loading a session (e.g. I used demo session) then it overwrites the session's search params in favour of the previously used params, which if you load the session from null is the last 24 hours.

This bug has been present for a while, e.g. on https://opsgateway-epac-dev.clf.stfc.ac.uk you can't see the bug as easily as the default is a null session, but if you set any search e.g. last 24 hrs timeframe, search, load a session and then add a filter, you'll see the search revert back to the search before you changed session.

I noticed this when testing the functions PR, as functions and filters work similarly. However, since this fix is generalised (ensuring we only run the useEffect which sets search params directly after `handleSearch`) this fix should fix the functions version of the bug as well.

It's quite hard to write unit tests for it, and it seemed too specific for an e2e test but I'm open to opinion on whether we should add a regression test.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Recreate the bug on e.g. dev and ensure the bug does not occur on this branch


joelvdavies (approved):
LGTM

joshdimanteto (approved):
LGTM

PR #449, opened 2024-09-06
DSEGOG-344 Upgrade to ESLint v9

joelvdavies:
## Description
Upgrades to eslint v9, removes an old CRA plugin. Uses flat config derived from IMS.

Fixes around 300 typescript errors in tests - on IMS we use `tsc --noEmit && tsc --noEmit -p cypress/tsconfig.json` as part of the linting step to include the unit and cypress tests in typescript checks, but I haven't copied that over as some tests would need to fundamentally change but I did use these and fixed the most obvious things I could find.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-344


joshdimanteto:
image
Warning in the units test

joshdimanteto (changes requested):
LGTM. just one minor error in the unit tests

joelvdavies:
> image Warning in the units test

Yeah I don't think that one has a very nice solution (its present in develop too), its the reason I decided against adding in those tests to IMS - its effectively due to the way the settings are imported before MSW is setup, so there is nothing to accept the request it sends.

joshdimanteto (dismissed):

joshdimanteto (approved):

louise-davies (approved):
LGTM 👍

PR #465, opened 2024-09-24
DSEGOG-347 add a favourite filter

joshdimanteto:
## Description
Add tab view for filters and favourite filters
Create add favourite filters dialog 
 
Going to do the e2e test with the real date in the last pr 
## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes DSEGOG-347


src/api/favouriteFilters.test.tsx
```diff
@@ -0,0 +1,38 @@
+import { renderHook, waitFor } from '@testing-library/react';
+import { FavouriteFilter } from '../app.types';
+import { hooksWrapperWithProviders } from '../testUtils';
+import { useAddFavouriteFilter } from './favouriteFilters';
+
+describe('session api functions', () => {
```
louise-davies: Probably needs to be named `favourite filters api functions`

src/filtering/favouriteFiltersDialogue.component.tsx
```diff
@@ -0,0 +1,138 @@
...
+  channels: Token[];
+}
+
+interface FavouriteFilterTokenised {
+  name: string;
+  filter: Token[];
+}
+
+interface FavouriteFilterError {
+  name?: string;
```
louise-davies, line 30: Why does the error have a name property? As far as I can see it's never used and makes the typing more complex when it could just be a string

src/views/dataView.component.test.tsx
```diff
@@ -59,14 +59,14 @@ describe('Data View', () => {
 
     await user.click(screen.getByRole('button', { name: 'Filters' }));
 
-    const dialogue = await screen.findByRole('dialog', { name: 'Filters' });
+    const dialogue = await screen.findByRole('dialog');
```
louise-davies: Why has this selector become less specific? The dialogue should have a name for a11y purposes and it makes testing like this easier...

src/views/viewTabs.component.tsx
```diff
@@ -44,7 +44,7 @@ function a11yProps(label: TabValue) {
   };
 }
 
-const StyledTab = styled(Tab)(({ theme }) => ({
+export const StyledTab = styled(Tab)(({ theme }) => ({
```
louise-davies, line 52: Since we're reusing the `StyledTab` code, it might be worthwhile trying to reuse the other code above - would just need to use generics to type `TabValue` I think

louise-davies (commented):
Just a few code comments but otherwise looks good :)

src/views/dataView.component.test.tsx
joshdimanteto: This shouldn't have been changed, I'll change it back.

src/filtering/favouriteFiltersDialogue.component.tsx
joshdimanteto: The name property is added for future-proofing, allowing for potential extensions like checking for duplicate names. While this hasn't been implemented in the backend yet, it could be useful as a user could create multiple filters with the same name, which doesn't look ideal in the frontend.

src/views/viewTabs.component.tsx
```diff
@@ -187,7 +200,7 @@ const ViewTabs = () => {
         <TabPanel value={value} label={'Data'}>
           <DataView sessionId={loadedSessionId} />
         </TabPanel>
-        <TabPanel value={value} label={'Plots'}>
+        <TabPanel<TabValue> value={value} label={'Plots'}>
```
joshdimanteto: image
image

`<TabValue>` is needed as it doesn't infer it from the value

louise-davies: Interesting, if I remove it then it still infers properly - and for example it's not on the `Data` tabpanel in the current commit so the type-checking must work on the CLI as the CI job is passing. If you want to check that the string literals are the correct type, you could do `'my-string' satisfies TabValue`. It's do we prefer adding the generic to the component, ensuring we use `satisfies` on all string literals or just being happy with `string` type being inferred and not checking against `TabValue`.

Also, I think there's a rogue `TabPanel` in `plotSettingsController.component.tsx` that can also use the new imports :)

joshdimanteto:
```
        <TabPanel value={value} label={value}>
          {value === 'Data' && <DataView sessionId={loadedSessionId} />}
          {value === 'Plots' && <PlotList />}
        </TabPanel>
```

or 
```
         const tabContent: Record<TabValue, React.ReactElement> = {
              Data: <DataView sessionId={loadedSessionId} />,
              Plots: <PlotList />,
            };

        <TabPanel value={value} label={value}>
          {tabContent[value]}
        </TabPanel>
```

if i refactor the code to look like this the type checking work correctly

louise-davies (approved):
LGTM

PR #479, opened 2024-10-03
Fix flaky e2e test by explicitly waiting for request before screenshotting

louise-davies:
## Description
When looking at renovate PRs, I noticed one of the unmocked e2e tests is extremely flaky. I (think I) fixed this by explicitly waiting for a request before trying to screenshot. On my machine, the test failed maybe 1/5 to 1/10 times, and I managed to get over 20 runs of the tests after my change to pass in a row, so it's probably fixed it I think!

I had to change the beforeEach code to select more than just the 1 image column to be able to differentiate between the table request and the image window thumbnail selector request, as otherwise they were identical but if we waited for the wrong one it would be flaky again...

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

### DIFF
--- a/e2e/real/images.spec.ts
+++ b/e2e/real/images.spec.ts
@@ -18,7 +18,7 @@ test.beforeEach(async ({ page }) => {
   await page.keyboard.press('ArrowDown');
   await page.keyboard.press('Enter');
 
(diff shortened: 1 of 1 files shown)


joelvdavies (approved):
Thanks for this, it was quite annoying 😄 - looks like its working

PR #490, opened 2024-10-15
DSEGOG-349 edit favourite filter

joshdimanteto:
## Description
A use should be able to edit a filter name or filter. There should be an error message if the user tries to edit a filter withoud chaning the values 
## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-349


louise-davies (commented):
Code looks good, one question is that I think I remember you talking about the cursor position in the filter inputs - does that get changed in a later PR? Because it does look odd to have the cursor at the front in the edit window, as well as the readonly filter inputs in the favourite filters tab let you change the position of the tokens when you click between tokens/use arrow keys after you click on them.

joshdimanteto:
'

> Code looks good, one question is that I think I remember you talking about the cursor position in the filter inputs - does that get changed in a later PR? Because it does look odd to have the cursor at the front in the edit window, as well as the readonly filter inputs in the favourite filters tab let you change the position of the tokens when you click between tokens/use arrow keys after you click on them.

Yeah the first thing mention has been changed in #511. Good spot in the readOnly mode, I didn't realised it did that :joy:. I have fix that in my most recent commit.

joshdimanteto:
image
image

louise-davies (approved):
LGTM - fixed the previous issue I raised

PR #511, opened 2024-11-04
DSEGOG-351 integrate favourite filters with filters

joshdimanteto:
## Description
- add favorite button
- add sections in autocomplete
- filter out duplicate filters  

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-351


louise-davies (changes requested):
Looks good overall, just some weirdness with the existing favourite filters... When I was filtering one of the favourite filters kept being duplicated. It seems as though the issue occurs when there's two _identical_ favourite filters (i.e the same tokens - occurs no matter whether they have the same name).

image

Probably should either prevent duplicate favourite filters, (and at that point probably also disallow duplicate names as that seems like it could get confusing for users), or figure out a way to make the autocomplete happier with duplicates...

joshdimanteto:
I think the best approach would be to prevent duplicate favourite filters and avoid duplicate names, as that could get confusing for users. I can add error validation to the frontend, but it would be better if this validation is also done on the backend. That way, even users interacting purely through the API can't input bad data.

louise-davies (approved):
LGTM - fixes the issue with duplicate filters being duplicated in the autocomplete and also prevents confusion between favourite filters with the same name :)

PR #513, opened 2024-11-06
DSGEOG-313 Add docker container for deployment

louise-davies:
## Description
Adds the Dockerfile to build a production docker container as well as the workflow to create containers on tags. See this workflow: https://github.com/ral-facilities/operationsgateway/actions/runs/11704416300/job/32596820832 and verify that the image is on Harbor to check.

Note I modified the "standard" frontend docker container to work when you supply the settings file via a volume as well as using env variables to edit a file. I feel as though the volume way will be needed for the other services as well, so figured I might as well experiment here.

I will shortly be adding a repo on Gitlab to detail the docker-compose file for deploying the service with the correct config etc. which I'll also link here to review that as well. EDIT: see the GitLab repo here with instructions: https://gitlab.stfc.ac.uk/scd-seg/operationsgateway-services/docker-operationsgateway

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-313


joshdimanteto:
Noted that you've modified the frontend Docker container to support both a settings file supplied via a volume and environment variables for configuration. Could you explain why both methods are necessary, and if certain scenarios specifically require a settings file over using only environment variables?

louise-davies:
@joshuadkitenge the environment variables I suppose are "easier" in that you only need to update those set variables and the rest is done for you, plus that's what is done currently in e.g. DataGateway & IMS. However, for e.g. SciGateway and probably DataGateway there's a lot more config than can be reasonably supported via variables, so I figured experimenting with a way to provide config files directly would be useful. For example, in OG, the `recordLimitWarning` option is not configured as an env variable and so would only currently be configurable via supplying a settings file. Again, we could just add every config option as an environment variable, but e.g. SG & DG will have more complex needs that will only be fixed by supplying via file so I figured it was worth testing on OG, keeping the old "simple" method which works when you're just spinning something up locally, but for production where you want to configure everything to just do that via a file rather than expose every config option as an env variable.

Plus the changes aren't that more complex - I just don't create the settings file in the image itself, and have a condition set up in the entrypoint script that tests if the file is supplied via volume before running the env variable code the other projects' images have

joshdimanteto:
@louise-davies  Thanks for the clarification

joshdimanteto (approved):
Tested locally in scigateway and it work well.

PR #521, opened 2024-11-28
DSEGOG-128 Image centroids/crosshairs

louise-davies:
## Description
Adds the ability to go into "crosshairs" mode. When switching into this mode, it will display the centroid for the image and the corresponding intensity plots. The user can click on the image to reposition the crosshairs. Resetting the view resets back to the centroid.

It should work when switching via the thumbnail selector, we always default to the centroid though.

image

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Play around with the crosshairs mode

## Agile board tracking
Closes DSEGOG-128


src/images/imagePlot.component.tsx
```diff
@@ -0,0 +1,213 @@
...
+            width: imageElement?.naturalWidth
+              ? imageElement.naturalWidth + 29
+              : 200,
+            height: 200,
+          }
+        : {
+            width: 200,
+            height: imageElement?.naturalHeight
+              ? imageElement.naturalHeight + 22
+              : 200,
```
joshdimanteto: Why does the natural width and height need 29 and 22 added ?

joshdimanteto (dismissed):
LGTM, work well just one question.

joshdimanteto:
Should an exported images also include the crosshairs ??

louise-davies:
@joshuadkitenge no it shouldn't (I think - they're not exported in eCat) - if they do it's an oversight on my part (although I wouldn't think they would as I would have thought we download the raw image itself, not the canvas overlay where the crosshairs are drawn...)

src/images/imagePlot.component.tsx
louise-davies: Ah, I should add a comment for this, but I needed to add some padding to account for the width/height of the Chart.js axis ticks (aka the numbers) so that the actual axis line itself aligns with the image.

Maybe I can try and calculate it based on the default config variables in Chart.js - but they may just have to remain magic numbers...

I will at least extract both out to variables at the top of the file (to make them easy to edit if we need) and explain what they're for

louise-davies:
@joshuadkitenge this is ready for review again as Stephen got back to me on how to handle 16-bit images. He picked the option where we let our plotting library autoscale the plot. See below example of a 16-bit image with intensity plots activated

image

joshdimanteto (commented):
LGTM to one minor thing. The plot on the right-hand side could do with a bit more spacing from the image so it either matches the spacing of the image underneath or is inline with the image control panel. 

image

louise-davies:
@joshuadkitenge I forgot to add spacing to the corresponding grid for the top row. I can't get it to align perfectly with the control panel (as the horizontal spacing of the right-hand graph kind of depends on the scale of the image pixel axis) but it aligns with the numbers on the other graph now

image

joshdimanteto:
That looks good.

joshdimanteto (approved):

PR #561, opened 2025-01-20
Error handling for OG

joshdimanteto:
## Description
- Create a ogApi axiosInstance
    - To manage 404 errors
    - set the authorisation headers
    - create handleOG_APIError to handle unknown errors
    - create the retryOG_APIErrors to hanlde retries
    - refactor all of the files in api directory to use ogApi

#### Reasoning 
- Implement the same method as IMS.
    - This method use the Axios interceptions. This method prevent multiple calls to re fetch the token if 
      multiple queries or mutations are called

- Retries are set to 3 attempts and it doesn't retry for 403 request as that is handle in the interceptors    

- Global error handler function set in the App.tsx handles any other errors 

- Reduces the amount code as the authorization header and api url are now set in the axios instance 

- Works well with the auto-refresh functionality currently implemented, as the automatic refresh is set to 1 minute.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-359


src/handleOG_APIError.ts
```diff
@@ -0,0 +1,44 @@
...
+  // due to the plugin its routing from being different). It is assumed that errors of this nature
+  // should not be possible due to SciGateway verifying the user itself, so we follow DataGateway's
+  // approach and don't display any of these errors.
+  if (broadcast && status !== 403) {
+    let broadcastMessage;
+    if (!error.response)
+      // No response so it's a network error
+      broadcastMessage =
+        'Network Error, please reload the page or try again later';
+    else if (status === 500)
```
louise-davies: We might want to check for all flavours of 5xx errors here rather than just specifically 500

src/api/sessions.tsx
```diff
@@ -70,12 +57,11 @@ export const useEditSession = (): UseMutationResult<
...
   SessionResponse
 > => {
-  const { apiUrl } = useAppSelector(selectUrls);
   const queryClient = useQueryClient();
   return useMutation({
-    mutationFn: (session: SessionResponse) => editSession(apiUrl, session),
+    mutationFn: (session: SessionResponse) => editSession(session),
     onError: (error) => {
-      console.log('Got error ' + error.message);
+      handleOG_APIError(error);
```
louise-davies: Most of the other mutations no longer have an `onError` handler in favour of handling it where the mutation is called. What's different about the `useEditSession`?

louise-davies (commented):
Seems to be working well, I was able to trigger some errors using a mocking extension and it correctly caused SG to either display error toasts or prompt a token revalidation. Just a couple minor comments :)

src/api/sessions.tsx
joshdimanteto: This is because all other mutations use `mutateAsync`, so the `handleOG_APIError` needs to be caught in the `catch` block. However, the `useEditSession` mutation uses `mutate` instead of `mutateAsync`. For error handling to work, the `handleOG_APIError` needs to be placed in the `onError` option of `useEditSession`.

louise-davies: Would it make sense for `useEditSession` to use `mutateAsync` as well then? Or we can leave it different, just wondering

louise-davies (approved):
LGTM 👍

PR #564, opened 2025-01-31
Refactor plots to use plotly.js instead of chart.js (part of DSEGOG-57)

louise-davies:
## Description
Motivation for this change is DSEGOG-57 - Plotly has the `rangebreaks` feature which can implement the skip non-working hours feature. That change will be in a follow up PR to try and keep this PR smaller as it's already a behemoth!

Overall the plots _should_ work identically to before - the only difference I can note is that we can't easily properly limit the zoom/pan range to the range of the data anymore. But since that only affects zoom/panning I decided it's not that big of a deal as the user can just choose to keep within bounds or not.

I had to rearrange some of the styling in the windows to accomodate Plotly to ensure the plot is correctly responsive and fills up the container as it is supposed to.

Also, some minor improvements in the playwright tests - we now use a better screenshot method (it takes two screenshots and if they're not the same it will continue taking subsequent screenshots until the page is stable before comparing). Also changed it so that Webkit screenshots are no longer huge - this is due to the default DPI of Webkit being 2 instead of 1.

## Testing instructions
- [ ] Review code
- [ ] Check Actions build
- [ ] Test plotting, traces & image crosshairs functionality to ensure migration has not lost any functionality

## Agile board tracking
connected to DSEGOG-57


e2e/mocked/plotting.spec.ts-snapshots/user-can-customize-both-left-and-right-y-axis-labels-1-chromium-linux.png
joshdimanteto, line 1: There is an extra point for the channel_DEFGH. this this expected

e2e/mocked/plotting.spec.ts-snapshots/user-can-customize-both-left-and-right-y-axis-labels-1-firefox-linux.png
joshdimanteto, line 1: same as above

e2e/mocked/plotting.spec.ts-snapshots/user-can-customize-both-left-and-right-y-axis-labels-1-webkit-linux.png
joshdimanteto, line 1: same as above

joshdimanteto (commented):
Works really well. Just some minor comments





 
I expected the data to be cleared when the user selects the `XY` as this is the current functionality on dev.  

https://github.com/user-attachments/assets/048eb1a4-3229-420d-a51d-86774ada457a

### DEV


https://github.com/user-attachments/assets/4c35ef31-d679-4458-b612-c3b0ac363960

e2e/mocked/plotting.spec.ts-snapshots/user-can-customize-both-left-and-right-y-axis-labels-1-chromium-linux.png
louise-davies: I think that yes this is expected. I think in the previous screenshot the `CHANNEL_DEFGH` happens to be covered up by the dipped shot number point. Plotly has the `CHANNEL_DEFGH` on top on shot number in comparison to Chart.js having shot number on top. and the slight differences in proportions of the plots is causing the plotly ones to not overlap quite as much.

louise-davies:
@joshuadkitenge good catch! Should be fixed now - I wasn't handling the case of `XAxis` being `undefined` properly.

joshdimanteto (dismissed):
LGTM

PR #565, opened 2025-02-05
force refresh being enabled and reduce rate

MatteoGuarnaccia5:
## Description

Forces autorefresh to be enabled.
Sets refresh rate to 30s


MatteoGuarnaccia5:
Ignore this PR, it was meant for a forked repo for the apprentice open evening. It somehow ended up on here as well

PR #569, opened 2025-02-05
DSEGOG-57 skip non-working hours toggle for plots

louise-davies:
## Description
This is the whole reason for the plotly refactor in #564 - to get access to Plotly's `rangebreak` functionality to implement a "skip non-business hours" toggle. I made the hours configurable in the settings file but opted to not let the days be configurable as I doubt they'll need to redefine what the weekend is but they might have a different idea of what "working hours" are that we don't want to hardcode.

See it in action below:

https://github.com/user-attachments/assets/99b03b95-6bfb-49bf-bd6a-b2f02694036c

I also restyled a little of the plot settings panel to reduce spacing as we're adding another option which makes the options larger vertically. Also, added some tooltips to the Scatter/Line buttons as I think I remember someone being confused about what they were. Plus fixed the fact that our date pickers were Sun-Sat instead of Mon-Sun by setting the locale to `en-GB`.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Test skip non-business hours feature

## Agile board tracking
Closes DSEGOG-57


louise-davies:
The commit order is strange for some reason - probably because I rebased this branch to be up to date with the base branch....

EDIT: fixed with setting the commit date to the author date and force pushing my local branch again :)

joshdimanteto (dismissed):
Works well 😃  . LGTM

PR #616, opened 2025-03-04
Improve functions functionality

joshdimanteto:
## Description
Improve Function Handling and Query Performance

-  Allowed all functions to be removed when necessary.
-  Enabled functions to be exported for better reusability.
-  Optimized query performance by only including function channels in conditions if they exist in the table, preventing unnecessary slowdowns.
-  Added extra explanatory text in parentheses for improved clarity.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-370

### DIFF
--- a/src/api/export.tsx
+++ b/src/api/export.tsx
@@ -1,6 +1,11 @@
 import { useMutation, UseMutationResult } from '@tanstack/react-query';
 import { AxiosError } from 'axios';
-import { SearchParams, SortType, type APIFunctionState } from '../app.types';
+import {
+  SearchParams,
+  SortType,
+  timeChannelName,
+  type APIFunctionState,
+} from '../app.types';
 import handleOG_APIError from '../handleOG_APIError';
 import { useAppSelector } from '../state/hooks';
 import { selectQueryParams } from '../state/slices/searchSlice';
@@ -75,7 +80,7 @@ export const exportData = async (
       channel in staticChannels ? `metadata.${channel}` : `channels.${channel}`;
     queryParams.append('projection', key);
 
-    if (!(channel in staticChannels) && !is_function) {
+    if (channel !== timeChannelName && !is_function) {
       existsConditions.push({ [key]: { $exists: true } });
     }
   });
--- a/src/api/export.tsx
+++ b/src/api/export.tsx
@@ -1,6 +1,6 @@
 import { useMutation, UseMutationResult } from '@tanstack/react-query';
 import { AxiosError } from 'axios';
-import { SearchParams, SortType } from '../app.types';
+import { SearchParams, SortType, type APIFunctionState } from '../app.types';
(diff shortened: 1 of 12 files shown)


src/api/export.tsx
```diff
@@ -62,17 +63,42 @@ export const exportData = async (
...
+    });
+
     // API recognises projection values as metadata.key or channel.key
     // Therefore, we must construct the appropriate parameter
     const key =
       channel in staticChannels ? `metadata.${channel}` : `channels.${channel}`;
     queryParams.append('projection', key);
 
-    if (!(channel in staticChannels)) {
+    if (!(channel in staticChannels) && !is_function) {
```
louise-davies: Just a minor comment unrelated to your change - we need to change the `!(channel in staticChannels)` to `channel !== timeChannelName` - as e.g. shot number and active experiment can both be undefined (it's probably worth going through the EPAC data interchange format document and checking our typing is all correct as well at some point)

I think this issue is also in the records and plot records endpoint as well

src/api/records.test.tsx
```diff
@@ -317,56 +317,6 @@ describe('records api functions', () => {
...
-        '{"$and":[{"metadata.timestamp":{"$gte":"2022-01-01 00:00:00","$lte":"2022-01-02 00:00:00"}}],"$or":[{"channels.CHANNEL_1":{"$exists":true}},{"channels.CHANNEL_2":{"$exists":true}}]}'
-      );
-
-      expect(new URL(request.url).searchParams.toString()).toEqual(
-        params.toString()
-      );
-      expect(result.current.data).toEqual(recordsJson.length);
-    });
-
     it('can set search and filter params via the store', async () => {
```
louise-davies: Might be worth renaming this test to explicitly state it's testing function params as well

src/functions/functionsHelp.component.tsx
```diff
@@ -107,7 +108,12 @@ const FunctionsHelp = (props: FunctionsHelpProps) => {
         . The Wizard will suggest suitable options and indicate using a grey box
         when each item has been recognised.
       </Body>
-      <Heading>Functions and operators included </Heading>
+      <Heading>
+        Functions and operators included{' '}
+        <Typography variant="caption" component="span" fontWeight="normal">
+          (Hover over name for more information)
```
louise-davies: In my opinion, `Hover over function name for more information` is more explicit & helpful

louise-davies (changes requested):
Verified that you can now:
- Export a function
- Delete a lone function
- Added help text is visible

Just some small comments

src/api/records.test.tsx
joshdimanteto: This test doesn't have functions in it but it should to test that functions are omitted from the projection

louise-davies (approved):
LGTM

PR #620, opened 2025-03-07
Get development credentials from env file & update readme with setup instructions

louise-davies:
## Description
See title

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


joshdimanteto (commented):
Wouldn't this also need to be changed for the  real api E2E test? Since we currently use the dev backend to test our real API E2E tests locally, or are we going to have to change that manually  ?

louise-davies:
@joshdimanteto Good catch! Yes, we should probably ensure that the env variables are read in the tests - the default value which CI will see is the CI credentials and then yeah it uses your local env file for e2e tests.

joshdimanteto (approved):
LGTM

PR #634, opened 2025-03-13
Bit depths DSEGOG-375

joshdimanteto:
## Description
- The bit_depth is supplied in the channel metadata (records endpoint) this should be forwarded to the image api hooks so that the correct bit depth is used to fetch for the image 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-375


e2e/mocked/images.spec.ts
```diff
@@ -145,7 +145,7 @@ test('user can change the false colour parameters of an image', async ({
     },
   });
 
-  await expect(slider.nth(0)).toHaveValue(`${0.4 * 255}`);
+  await expect(slider.nth(0)).toHaveValue('1635');
```
louise-davies, line 148: Could this not continue to be calculated like `0.4 * 4095`?

e2e/real/images.spec.ts
```diff
@@ -143,6 +143,137 @@ test('user can change the false colour parameters of an image', async ({
...
 
+test('user can change the false colour parameters of an 12 bit image', async ({
+  page,
+}) => {
+  await page.getByLabel('to, date-time input').fill('2023-06-06 12:10');
+  await page.getByLabel('from, date-time input').fill('2023-06-06 12:00');
+
+  await page.getByRole('button', { name: 'Search', exact: true }).click();
+
+  // add trace channel to the table so we can click on a trace
```
louise-davies:
```suggestion
  // add image channel to the table so we can click on an image
```

louise-davies (changes requested):
This works fine when clicking an image from the table, but when using the thumbnail selector it does not change bit depth correctly. Need to also edit `windows/ThumbnailSelector` to pass the potential new bit depth alongside the new record id.

e2e/mocked/images.spec.ts
joshdimanteto: I did try that initially, but that value gives '1638' instead of '1635', causing the test to fail

louise-davies: That's odd that it's no longer exact anymore oh well

joshdimanteto:
> This works fine when clicking an image from the table, but when using the thumbnail selector it does not change bit depth correctly. Need to also edit `windows/ThumbnailSelector` to pass the potential new bit depth alongside the new record id.

Good spot. I've addressed this in the latest commit

louise-davies (approved):
LGTM

PR #642, opened 2025-03-19
Add Menu item and dialog to export a images or waveforms column DSEGOG-122

joshdimanteto:
## Description
- disable sort for functions and images and waveforms

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-122


src/export/exportChannelColumn.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <>
+      <Dialog open={open} onClose={onClose} maxWidth="lg">
+        <DialogTitle>Export Channel</DialogTitle>
+        <DialogContent>
+          Do you want to export all files in the{' '}
+          <strong data-testid="export-channel-name">
+            {channelInfo.name || channelInfo.systemName}
+          </strong>
+          ?
```
louise-davies: This reads a little clunky - need to either add `column` to the end after the channel name, or rephrease to `Do you want to export all files for the channel X`

src/export/exportChannelColumn.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { mutateAsync: exportChannels, isPending } = useExportData();
+
+  const handleExportChannel = React.useCallback(() => {
+    exportChannels({
+      exportType: 'All Rows',
+      dataToExport: {
+        Scalars: true,
+        Images: true,
+        'Waveform CSVs': true,
+        'Waveform Images': true,
```
louise-davies: Might need to check if we need this - I'd go with it set to false for now as that is the behaviour of `eCat`

louise-davies (changes requested):
Works well for me, and the backend is correctly appending the column name to the filename. Just a few minor comments

src/export/exportChannelColumn.component.tsx
joshdimanteto: is that just for waveform Images ?

joshdimanteto: If so, I'll also set scalar to false, as this dialog only exports image and waveform channels.

louise-davies: Yeah, just for waveform thumbnails as in `eCat` just exports the CSVs

louise-davies (commented):
Exporting a single column works well. Just a thing I noticed - if we have functions that are defined but not visible in the table, we're still sending the function state to the backend which is slowing down e.g. single column exports. You had done some work to stop projecting on functions that are not in the table, but maybe we need to think more about including the functions in the API request at all if they're not in use.

At the very least, we should never include functions in a single column export unless we're explicitly exporting that function as the function data is never needed I don't think

joshdimanteto:
> Exporting a single column works well. Just a thing I noticed - if we have functions that are defined but not visible in the table, we're still sending the function state to the backend which is slowing down e.g. single column exports. You had done some work to stop projecting on functions that are not in the table, but maybe we need to think more about including the functions in the API request at all if they're not in use.
> 
> At the very least, we should never include functions in a single column export unless we're explicitly exporting that function as the function data is never needed I don't think

The reason why those functions are sent as part of the request is that functions can be dependent on each other. For example, function 1: `test1 = 1 + channel_X`, and function 2: `test2 = test1 + channel_Y`. So, the API needs to know what function `test1` to generate the value for `test2`.

joshdimanteto:
It is still should be possible to optimize by having recursion check for dependent functions and conditional adding them in.

src/api/functions.tsx
```diff
@@ -61,21 +73,20 @@ export function convertExpressionsToStrings(
...
+        channels: [...(channels ?? []), ...depChannels],
+        functions: [name, ...depFunctions],
       };
     }
   );
 
   return {
     functions,
-    functionsWithChannels,
+    functionsWithDeps: functionsWithChannels,
```
louise-davies:
```suggestion
  const functionsWithDeps = functionStates.map(
    ({ name, expression, channels }) => {
      const { functions: depFunctions, channels: depChannels } =
        getDepsRecursively(expression, functionStates);
      return {
        name,
        expression: transformExpression(expression),
        channels: [...(channels ?? []), ...depChannels],
        functions: [name, ...depFunctions],
      };
    }
  );

  return {
    functions,
    functionsWithDeps,
```

Very minor comment but probably makes sense to rename this as well, also makes the return simpler as can go back to the shorthand

src/api/records.tsx
```diff
@@ -101,7 +101,7 @@ const fetchRecords = async (
   });
```
louise-davies, line 108: I can't comment on the above - but I thought the whole point of the previous commit was to not send functions if they're not visible in the table? Yes we're now not projecting on them, but we're still sending the functions to the backend in the above and this still slows down the queries

louise-davies: Ah, just realised that it was fixed in `export.tsx` but not in `records.tsx` - just need the same logic as exports I think

louise-davies (approved):
LGTM

PR #655, opened 2025-03-25
vector bar chart channel DSEGOG-400

joshdimanteto:
## Description
Enter a description of the changes here

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-400


joshdimanteto:
image

joshdimanteto:
image

e2e/mocked/vectors.spec.ts-snapshots/user-can-limit-the-vector-data-2-chromium-linux.png
louise-davies, line 1: The `user can limit the vector data` screenshots don't differ between 1 and 2 - is this intentional or is it a holdover from e.g. plotting tests? The test clicks the reset button - is that supposed to do anything here?

src/api/channels.tsx
```diff
@@ -150,9 +155,10 @@ export const constructColumnDefs = (
         return <React.Fragment>{wordWrap.join('')}</React.Fragment>;
       },
       meta: { channelInfo: channel },
-      cell: isChannelMetadataScalar(channel)
-        ? ({ getValue }) => {
-            const value = getValue();
+      cell: ({ row, getValue }) => {
+        const value = getValue<string>();
```
louise-davies: Technically this type is wrong in the scalar case which can be a number. Not sure this really matters though...

src/state/slices/windowSlice.test.tsx
```diff
@@ -76,7 +77,50 @@ describe('windowSlice', () => {
...
+          type: 'float_image',
+          recordId: '1',
+          channelName: 'TEST',
+          title: 'Image TEST 1',
+          ...DEFAULT_WINDOW_VARS,
+        } satisfies WindowConfigType,
+      });
+    });
+
+    it('openVectorWindow creates a float image with the default options', () => {
```
louise-davies:
```suggestion
    it('openVectorWindow creates a vector with the default options', () => {
```

src/windows/windowPortal.component.tsx
```diff
@@ -58,7 +58,7 @@ export default class WindowPortal extends React.PureComponent<
       // we do this so that all the Plotly.js code which relies on window references the correct window (i.e. the popup)
       const plotlyjsScript = document.createElement('script');
       plotlyjsScript.src =
-        'https://cdnjs.cloudflare.com/ajax/libs/plotly.js/2.35.3/plotly-gl2d.min.js'; // need the gl2d bundle so include scattergl for traces optimisation
+        'https://cdnjs.cloudflare.com/ajax/libs/plotly.js/2.35.3/plotly.min.js'; // need the main bundle so include scattergl for traces optimisation and bars charts for vectors
```
louise-davies:
```suggestion
        'https://cdnjs.cloudflare.com/ajax/libs/plotly.js/2.35.3/plotly.min.js'; // need the main bundle to include scattergl for traces optimisation and bars charts for vectors
```

louise-davies (commented):
Seems to be working well with both mock data and real API. Only minor overall comment is that the width of the slider makes it so the actual plot is very small when the window is small - could we explore either making this smaller or allowing it to be hidden? Equally, as this is one of the areas where we don't have wireframes, we could leave it for now and wait for feedback from users.

joshdimanteto:
image

joshdimanteto:
@louise-davies I think I should change the display names from "Vector Skip" and "Vector Limit" to "Lower Bound" and "Upper Bound", as those are more descriptive. What are your thoughts on this?

louise-davies:
@joshdimanteto that sounds good - it also aligns more with the image colourmap controls as well. I think we go with lower and upper bounds and if EPAC want different names then they can let us know

joshdimanteto:
@louise-davies I don't think there are any changes to make in this PR, as the "Vector Limit" and "Vector Skip" labels were added in the next one (https://github.com/ral-facilities/operationsgateway/pull/659). So this is ready for review.

src/vectors/vectorWindow.component.tsx
```diff
@@ -0,0 +1,211 @@
...
+              wrap="nowrap"
+              spacing={1}
+            >
+              <Grid
+                item
+                m={1}
+                xs
+                sx={{
+                  height: '100%',
+                  minWidth: '80%',
```
louise-davies: I don't like this - this often causes overflows/overlaps. I think instead we just remove the `width: 300` and `margin: 1` from `vectorControlsPanel` to make that element smaller as I think it's a little wide atm.

joshdimanteto: The reason I chose width: 300px is because it matches the image control panel, and the sliders in both windows should be the same width. I also ignored the overflow and overlap issues, as the same thing happens in the image window if an image is too large.
image .

louise-davies: yeah, but the difference is the image is a fixed size whereas the graph here grows/shrinks to fill the remaining space. Plus the image control panel has more controls in it that actually fill the width (e.g. the colour map selector needs to be quite wide as some of the colour maps have long names etc.)

That's just my preference though - I think I'd rather it match more closely with the other two plotting windows in that the graph fills the whole window without overflow rather than it matching the image window due to similar controls

joshdimanteto: I did try to make the control panel optional—so it would be closed initially and opened by clicking a "Show Vector Control Panel" button located in the vector window’s controls—but I had trouble resizing the plot when the panel was conditionally rendered. The plot only resizes when the window itself is resized, not when the DOM changes.

louise-davies: Hmm, the main plotting window manages to have the plot resize when the settings gets hidden/shown - must be something specific that is needed for it to be able to detect that the graph div has resized. Also, I think there is a manual `Plotly.Plots.resize(graphDiv)` function but I'm unsure if it's useful

joshdimanteto: I managed to get it working by conditional rendering the control panel.

joshdimanteto: @louise-davies 



https://github.com/user-attachments/assets/ec2d0c71-15a1-4884-b3ee-be9c929f4782

louise-davies: That looks good :) still personally think the panel could be smaller myself, but happy to keep it as is now it's on a toggle so users can hide it when not using it

src/vectors/vectorWindow.component.test.tsx
```diff
@@ -65,6 +81,41 @@ describe('Vector Window component', () => {
...
+      />
+    );
+
+    await user.click(
+      screen.getByRole('button', { name: 'Show Vector Controls' })
+    );
+    expect(
+      await screen.findByRole('button', { name: 'Hide Vector Controls' })
+    ).toBeVisible();
+
```
louise-davies, line 116: Might be worth making the `vi.fn()` external to the `Object.defineProperty` portion so we can assert that `resize` has been called here?

louise-davies (commented):
Other than my minor testing comment above, all looks good to me :)

louise-davies (dismissed):

louise-davies (approved):
LGTM

PR #660, opened 2025-03-31
fix small bugs DSEGOG-401 DSEGOG-404 DSEGOG-413

joshdimanteto:
## Description
make the record range converter a slient error 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
closes DSEGOG-401, closes DSEGOG-404, closes DSEGOG-413

### DIFF
--- a/src/search/components/dateTime.component.tsx
+++ b/src/search/components/dateTime.component.tsx
@@ -295,6 +295,9 @@ const DateTimeSearch = (props: DateTimeSearchProps): React.ReactElement => {
             <DateTimePicker
               format="yyyy-MM-dd HH:mm"
               value={datePickerFromDate}
+              referenceDate={
+                datePickerFromDate || datePickerToDate || new Date()
+              }
               maxDateTime={datePickerToDate || new Date('2100-01-01 00:00:00')}
               onChange={(date) => {
                 setDatePickerFromDate(date);
@@ -396,6 +399,9 @@ const DateTimeSearch = (props: DateTimeSearchProps): React.ReactElement => {
             <DateTimePicker
               format="yyyy-MM-dd HH:mm"
               value={datePickerToDate}
+              referenceDate={
+                datePickerToDate || datePickerFromDate || new Date()
+              }
               minDateTime={
                 datePickerFromDate || new Date('1984-01-01 00:00:00')
               }
--- a/src/search/components/dateTime.component.tsx
+++ b/src/search/components/dateTime.component.tsx
@@ -12,6 +12,7 @@ import {
   DateTimePicker,
   DateTimeValidationError,
   LocalizationProvider,
+  pickersLayoutClasses,
 } from '@mui/x-date-pickers';
 import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFnsV3';
 import { PickersDay, PickersDayProps } from '@mui/x-date-pickers/PickersDay';
@@ -295,9 +296,6 @@ const DateTimeSearch = (props: DateTimeSearchProps): React.ReactElement => {
             <DateTimePicker
               format="yyyy-MM-dd HH:mm"
               value={datePickerFromDate}
-              referenceDate={
-                datePickerFromDate || datePickerToDate || new Date()
-              }
               maxDateTime={datePickerToDate || new Date('2100-01-01 00:00:00')}
               onChange={(date) => {
                 setDatePickerFromDate(date);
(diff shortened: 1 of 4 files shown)


src/search/searchBar.component.tsx
```diff
@@ -322,14 +322,7 @@ const SearchBar = (props: SearchBarProps): React.ReactElement => {
...
     setParamsUpdated(false);
-  }, [
-    dateRange.fromDate,
-    dateRange.toDate,
-    experimentID,
-    maxShotsParam,
-    shotnumRange.max,
-    shotnumRange.min,
-  ]);
+  }, [dateRange, experimentID, maxShotsParam, shotnumRange]);
```
joshdimanteto, line 325: The second useEffect works better because it tracks the entire object, ensuring the effect runs whenever the object itself changes, whereas the first only tracks specific properties, potentially missing updates when the object is replaced.

src/traces/tracePlot.component.tsx
```diff
@@ -54,6 +54,12 @@ const TracePlot = (props: TracePlotProps) => {
           type: 'linear',
           exponentformat: 'none',
           automargin: true,
+          tickformat: '.4g',
```
joshdimanteto: The tickformat: '.4g' ensures that x-axis values are displayed with a maximum of 4 significant figures. The tickformatstops allow finer control over precision in different numerical ranges:


-    [0, 0.01] → Uses '.2g' for higher precision in very small values.
-    [0.01, 1] → Uses '.3g' for moderate precision in small values.
-    [1, null] → Uses '.4g' to maintain standard formatting for larger values.
-    This approach ensures clarity while avoiding excessive decimal places or scientific notation where unnecessary.

joshdimanteto: Change it to a global one, remove tick format.

joshdimanteto: remove and put in a separate pr

joshdimanteto: this has been removed from this pr

src/search/components/dateTime.component.tsx
```diff
@@ -295,6 +295,9 @@ const DateTimeSearch = (props: DateTimeSearchProps): React.ReactElement => {
             <DateTimePicker
               format="yyyy-MM-dd HH:mm"
               value={datePickerFromDate}
+              referenceDate={
+                datePickerFromDate || datePickerToDate || new Date()
+              }
```
joshdimanteto: This changes means that if the fromDate is null that the date picker is opened on the ToDate, if that is also null then it gets opened on the current date

joshdimanteto: Users are still are still able to navigate to today's date by clicking the today button in the date picker

louise-davies (changes requested):
If we add this code to the `slotProps` (just need to switch from date to to date and vice versa) then we add a custom button to the bottom that allows going to the other date even when a different date is already filled in:

```
layout: {
  sx: {
    [`.${pickersLayoutClasses.shortcuts}`]: {
      gridColumn: 2,
      gridRow: 3,
    },
  },
},
shortcuts: {
  items: [
    {
      label: 'Jump to from date',
      getValue: () => {
        return searchParameterFromDate;
      },
    },
  ],
  changeImportance: 'set',
},
```

Thoughts? I think it's a nice addition and relatively simple code to maintain as MUI has it easily configurable

joshdimanteto:
> If we add this code to the `slotProps` (just need to switch from date to to date and vice versa) then we add a custom button to the bottom that allows going to the other date even when a different date is already filled in:
> 
> ```
> layout: {
>   sx: {
>     [`.${pickersLayoutClasses.shortcuts}`]: {
>       gridColumn: 2,
>       gridRow: 3,
>     },
>   },
> },
> shortcuts: {
>   items: [
>     {
>       label: 'Jump to from date',
>       getValue: () => {
>         return searchParameterFromDate;
>       },
>     },
>   ],
>   changeImportance: 'set',
> },
> ```
> 
> Thoughts? I think it's a nice addition and relatively simple code to maintain as MUI has it easily configurable

I also looked into using the custom action bar so that the action buttons in the date picker would have the same styling.
image

However, since the bottom buttons — "Clear" and "Today" — close the dialog when clicked, it's better for them not to look the same as buttons with different functionality.

I have also remove the referenced date as the user now has a way to easily navigate to the from or to date on the date pickers

I've implemented the shortcuts method in the latest commit.

louise-davies (approved):
LGTM

PR #680, opened 2025-04-16
DSEGOG-405 fix names of active_area & active_experiment

louise-davies:
## Description
They were previously camelCase when they needed to be snake_case. Also made it so that we type `active_area` correctly as required in the metadata (as this is what the data interchange format doc says it is) & thus added `active_area` to all the records in the mock, plus added a single record with an `active_experiment`

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] See Jira issue for specific record (shot number 402480061800) that should now display both active area & active experiment

## Agile board tracking
Closes DSEGOG-405


joshdimanteto (approved):
LGTM, works as expected on this branch

PR #691, opened 2025-05-15
DSEGOG-346 upgrade to MUIv6

joelvdavies:
## Description
Upgrades to MUI v6. MRT v3 doesn't mention it adds support to MUI v6, but instead that MUI v6 is the new minimum for that version so have not upgraded here as it doesn't seem required.

The required code changes for the intiial upgrade were minimal, I made use of the applyStyles function for dark mode as its recommended to reduce flickering when changing themes, though we still use mode to manually modify the theme's blue colour and would require changes to scigateway to eliminate properly.

I ended up temporarily enabling https://typescript-eslint.io/getting-started/typed-linting/ to find some deprecated code, and fixed a few other things it noticed e.g. some async functions missing awaits. I also updated all grids to use v2, (largely just props changes, some minor changes to spacing can be observed in some places).

The only outstand deprecated code I am aware of is the `CustomTextField` component present in both the `dateTime.component.tsx` and `xAxisTab.component.tsx` as they still use `inputProps`. I tried refactoring this but at present the anchor ref is still in input props, and due to the extra custom parameters it seems when I wrap it in a function or separate component in order to allow for these extra props, the anchor breaks when closing/opening so the menu appears in the wrong place. So I ended up leaving this to MUI v7/datepickers v8.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build

## Agile board tracking
Closes DSEGOG-346, Closes DSEGOG-421


package.json
```diff
@@ -10,10 +10,10 @@
     "@emotion/styled": "11.14.0",
     "@hello-pangea/dnd": "18.0.1",
     "@hookform/resolvers": "^4.0.0",
-    "@mui/base": "5.0.0-beta.40",
-    "@mui/icons-material": "5.17.1",
-    "@mui/material": "5.17.1",
-    "@mui/system": "5.17.1",
+    "@mui/base": "5.0.0-beta.70",
```
louise-davies: Isn't this package now deprecated? I think we're only using it for `ClickAwayListener` - which previously they pointed us to import from `@mui/base` but now the docs say to import it straight from `@mui/material`: https://v6.mui.com/material-ui/react-click-away-listener/

src/table/table.component.tsx
```diff
@@ -435,12 +437,14 @@ const Table = React.memo((props: TableProps): React.ReactElement => {
...
-            {/* eslint-disable-next-line jsx-a11y/role-supports-aria-props */}
+            {}
             <Backdrop
               component="tr"
+              // At the moment component isn't working: https://github.com/mui/material-ui/issues/44283
+              slots={{ root: 'tr' }}
               sx={{ position: 'absolute', zIndex: 100, height: 'inherit' }}
               open={!loadedData}
-              role="none"
+              // role="none"
```
louise-davies: Why the change to comment out the `role="none"`? I presume this was needed at some point to ensure that it counted as a valid HTML table without messing up the accessibility roles. If this is fine now - remove the `role="none"` comment and also the empty `{}`?

One thing I note is that the backdrop no longer looks like a backdrop - instead it is just the loading spinner. Also our custom styles aren't applying due to this workaround. I think we should try and get the MUI issue re-opened/open a new issue by providing a minimal example.

I don't think this is a blocking issue as the spinner still works, just doesn't look as nice as the previous one...

louise-davies (changes requested):
Overall looks good and mostly everything seems like it has migrated well except for the table loading backdrop.

package.json
joelvdavies: Yeah it is, thanks.

src/table/table.component.tsx
joelvdavies: The role="none" was commented out due to an eslint error `The attribute aria-hidden is not supported by the role none.eslint`[jsx-a11y/role-supports-aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/role-supports-aria-props.md) - But that was only intended as a temporary fix, I just forgot about it, I will address it.

I will also have a look at fixing the backdrop, looks like the component issue is still in v7 too, I cant see an obvious way to reopen the issue other than commenting on it, so I think I will create a new one.

joelvdavies: Changing the backdrops component seems to be what is causing it to no longer display correctly, Fast pass is not detecting any issue with the table even if I remove the tr and td there and just use the backdrop which seems to recreate the old behaviour. So does wrapping it inside a <td> and <tr> element so I have gone with that so it is still valid.

joelvdavies: https://github.com/mui/material-ui/issues/46264

joelvdavies:
As discussed for the progress bar, we could either go for a minimum height:

image

image


Or try and fill the page:
image

image

louise-davies (approved):
LGTM - the progress bar is working now as it's actually visible and the backdrop is working :)

joshdimanteto (approved):

PR #697, opened 2025-05-29
Add labels in channels for vectors

joshdimanteto:
## Description
- Updated app types to support channel labels
- Enhanced channel metadata panel to display and manage labels
- Conditionally x and y units in channel metadata panel
- add channel test for vectors and float_images

## Testing instructions

- [ ] Review code
- [ ] Check Actions build


joshdimanteto:
image

patrick-austin:
Note that there are two places that metadata fields including `units`, `x_units` and `y_units` can be defined. The first is in the [channel manifest](https://github.com/ral-facilities/operationsgateway-api/blob/main/test/endpoints/test_manifest.json). Whatever is defined here will be returned by requests to [/channels](https://github.com/ral-facilities/operationsgateway-api/blob/8955ad50b418149a575222f738347497c219ef7f/operationsgateway_api/src/routes/channels.py#L72-L89) or [/channels/{channel_system_name}](https://github.com/ral-facilities/operationsgateway-api/blob/8955ad50b418149a575222f738347497c219ef7f/operationsgateway_api/src/routes/channels.py#L92-L118).

The second is within the ingested hdf5 file corresponding to a single record. This metadata belongs to a specific record and a specific channel, and will be returned by requests to [/records](https://github.com/ral-facilities/operationsgateway-api/blob/8955ad50b418149a575222f738347497c219ef7f/operationsgateway_api/src/routes/records.py#L30-L144) and [/records/{_id}](https://github.com/ral-facilities/operationsgateway-api/blob/8955ad50b418149a575222f738347497c219ef7f/operationsgateway_api/src/routes/records.py#L198-L272).

These two sources of information do not necessarily match. In our test data, for example, none of the waveform channels have units in the manifest, but the records do have units. A reason for this is to account for the fact that instrumentation may be modified (resulting in different metadata) but record the same thing overall. For example, the same camera might be used but its settings adjusted to increase gain. In this case it would not be appropriate to assign metadata to the manifest, as it wouldn't apply to all records, but it does make sense for each record to track its own metadata.

At the end of the day, despite the similarity in naming and structure, units from the manifest are distinct from units from the record, and they may differ. All we can do is return and display the values that we have defined, whatever the reasoning behind it.

joshdimanteto:
image

louise-davies (approved):
LGTM

PR #708, opened 2025-06-27
Limit plot data using the search date range DSEGOG-425

joshdimanteto:
## Description
see title 

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-425


louise-davies (changes requested):
Not sure I like having these as just min/maxes - the user just sees a generic `Date-time format: yyyy-MM-dd HH:mm` error - plus if you manually type in a date outside the range it still updates the plot to that range. I think maybe we just add a "Jump to search range" (name pending) shortcut in these pickers as that is potentially helpful and isn't too intrusive either

joshdimanteto:
#718

PR #710, opened 2025-07-01
DSEGOG-422 save plots on session save

joshdimanteto:
## Description
see title

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] {more steps here}

## Agile board tracking
closes DSEGOG-422


e2e/mocked/sessions.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+    case 'webkit':
+      screenY = 0;
+      screenX = 0;
+      colour = '#40807e';
+      break;
+    default:
+      screenY = 200;
+      screenX = 200;
+      colour = '#41817f';
+  }
```
louise-davies: I wonder if instead of all this complexity of different browsers having different colours - to try and instead do a partial match (e.g. using `toMatchObject` instead of `toEqual` which allows for matching a subset of properties) so we only check for session properties we care about validating or using `expect.objectContaining()` and using e.g. `expect.any(string)` on values that we don't care about.

Also, why are we changing the plot colour in this test anyway? The test is the same if we just select the channel without selecting the colour or changing something easier e.g. the title of the plot.

e2e/mocked/sessions.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+      functions: { appliedFunctions: [] },
+      plots: {
+        [plotId]: {
+          XAxis: 'timestamp',
+          XAxisScale: 'date',
+          axesLabelsVisible: true,
+          gridVisible: true,
+          id: plotId,
+          innerHeight: 400,
+          innerWidth: 600,
```
louise-davies: I think I remember when looking at this when I implemented the saving positions code that there's no way to resize the popup in playwright? I'm not sure if this is still true, if we can resize the window it would be useful to test that here as well

louise-davies:
Add extra e2e test for non-plot windows saving their position/size

louise-davies (approved):
LGTM

PR #728, opened 2025-09-22
DSEGOG-442 Always assume API uses UTC timestamps 

louise-davies:
## Description
Ensure that we only send UTC timestamps to the API (needed an upgrade of date-fns to v4 to use their new timezone support to format it nicely) and ensure we always parse API timestamps as UTC - to ensure this I've tried to make a centralised util function convertAPITimestampToDate to handle converting an API timestamp string to a JS Date. To test this, you'll need to look at the requests to see that currently because we're in BST all timestamps sent/received from the API are 1 hour earlier than displayed. Whereas all displayed timestamps should be in the current time/BST.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Check against preprod to check API handling

## Agile board tracking
Closes DSEGOG-442

### DIFF
--- a/e2e/mocked/vectors.spec.ts
+++ b/e2e/mocked/vectors.spec.ts
@@ -345,6 +345,8 @@ test('should display an error if vector upper bound or skip is not a valid numbe
 
   const vectorLimitInput = await page.getByLabel('Upper Bound');
 
+  await vectorLimitInput.scrollIntoViewIfNeeded();
+
   await vectorLimitInput.fill('abc');
 
   await expect(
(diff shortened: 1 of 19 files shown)


joshdimanteto (dismissed):
Works well. Just one minor comment: in the handlers file, we intercept and mock the experiment start and end dates on line 160-204. I think the date format should be updated to match the API format currently it is using ISO.

Now that I'm looking at this again, I don't think this code is used anymore. It was originally using MSW to mock the experiment date ranges for the real API data. However, MSW is no longer used at all when the frontend interacts with the actual API (in both e2e tests and dev mode). This code should probably be removed, but that can be done outside this PR.

louise-davies:
@joshdimanteto yes, technically if that mock code is we could have timezones issues but as you said I don't think that code is used any more. It may just be easier to delete the code in this PR so we don't forget, as as you said it's not relied on in any of the test code so I just need to remove the `*/experiments` handler

joshdimanteto (approved):

PR #729, opened 2025-09-25
DSEGOG-443 Fix refresh bug

louise-davies:
## Description
The refresh basically never worked when the minute changed as the shot number -> date conversion would take priority and wipe out the timeframe. So added a flag state variable to detect when we're in the special "time frame has updated after a refresh state" to disable the shot number -> date range useEffect until the date -> shot number useEffect runs. Also delay committing the search until after the useEffect, this is so that the committed search has the correct shot numbers.

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [x] Test auto-refresh against prod data (as that has proper shot numbers and recent data to test)
- [x] Test auto-refresh behaves OK if shot number conversion fails (the search request sent a little late in this case due to react-query auto-retries and back-offs, but this error should be rarer in production I hope... plus eventually it updates the search successfully. Only way to stop this delay would be to stop retries/reduce backoff for the conversion requests)

## Agile board tracking
Closes DSEGOG-443

### DIFF
--- a/cypress/e2e/search.cy.ts
+++ b/cypress/e2e/search.cy.ts
@@ -298,6 +298,17 @@ describe('Search', () => {
       });
 
       it('refreshes datetime stamps and launches search if timeframe is set and refresh button clicked', () => {
+        cy.window().then(async (window) => {
+          // Reference global instances set in "src/mocks/browser.js".
+          const { worker, http } = window.msw;
+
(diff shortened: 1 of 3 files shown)


joshdimanteto (approved):
LGTM, work well and tested with PROD and DEV APIs

PR #731, opened 2025-09-29
DSEGOG-466 add unsaved data prompt before plot window closes

louise-davies:
## Description
Utilise the [beforeunload event](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) dialogue to present the user with a "you have unsaved changes" dialogue before actually closing a plotting window if the current plotting config is different to what is stored in Redux. We already hooked into onBeforeUnload, so move those handlers to onClose to allow for an optional onBeforeUnload handler in windowPortal.

Also, realised that we no longer save window positions when using the plot save button which seems wrong so I added that back. But that would mean that resizing the window would prompt the "unsaved changes" dialogue which I believe is a bit overkill, so I excluded the window params from the equality check to see if there's unsaved changes.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Test that trying to close a plotting window with unsaved changes prompts with a are you sure dialogue (note you have to have interacted with the window in some way, see [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes))
- [x] Test that trying to close a plotting window with no unsaved changes just closes without prompting.

## Agile board tracking
Closes DSEGOG-466

### DIFF
--- a/e2e/mocked/plotting.spec.ts
+++ b/e2e/mocked/plotting.spec.ts
@@ -1,4 +1,5 @@
 import { expect, test } from '@playwright/test';
+import assert from 'node:assert';
 import type { PlotlyHTMLElement } from 'plotly.js';
 
 test('plots a time vs shotnum graph and change the plot colour', async ({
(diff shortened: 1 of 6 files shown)


louise-davies:
note that we'll have to ignore the patch check here as we cannot test the `onBeforeClose` handler in `plottingWindow` via unit tests

joshdimanteto (approved):
LGTM

PR #776, opened 2026-02-27
DSEGOG-526 Fix reversed trace plots

louise-davies:
## Description
Switching waveforms via the thumbnail selector could cause either the x or y axis to invert, but would be fixed by clicking the reset view button. The cause of this was passing in null min/max parameters when the waveform is undefined whilst the new waveform is loading. Explicitly checking for empty data and not setting min/max parameters fixes the issue

## Testing instructions

- [ ] Review code
- [ ] Check Actions build
- [ ] Use demo session on OG dev, channel FE-204-NSS-SP (stretcher output spectrum), on OG dev this channel is extremely prone to inverting on both axes. Check that running this branch and switching between waveforms via the thumbnail selector on the same channel doesn't cause any inversions. Note that data needs to be fetched from the API to cause possible inversion, cached data doesn't have this issue

## Agile board tracking
Closes DSEGOG-526


src/images/imagePlot.component.tsx
```diff
@@ -150,7 +150,7 @@ const ImagePlot = (
     palette: { mode: themeMode },
   } = useTheme();
 
-  const [optionsString, setOptionsString] = React.useState(
+  const [layoutStrong, setLayoutString] = React.useState(
```
joshdimanteto:
```suggestion
  const [layoutString, setLayoutString] = React.useState(
```

src/images/imagePlot.component.tsx
```diff
@@ -237,15 +237,15 @@ const ImagePlot = (
...
+    setLayoutString(JSON.stringify(newChartOptions));
   }, [chartOptions, crosshairPosition, imageDims, themeMode, type]);
 
   /* This canvas is turned into a Plotly.js plot via code in windowPortal.component.tsx */
   return (
     <Box
       className="plotly-chart"
       data-config={JSON.stringify(plotlyConfig)}
-      data-layout={optionsString}
+      data-layout={layoutStrong}
```
joshdimanteto:
```suggestion
      data-layout={layoutString}
```

joshdimanteto (changes requested):
works well, just one minor comment

joshdimanteto (approved):
LGTM

PR #838, opened 2026-04-24
DSEGOG-530 DSEGOG-528 data types

louise-davies:
## Description
Add the ability to filter by data types - this is enabled by providing a non-empty array `dataTypes` in the settings. When this is provided, a few things happen:

- data type selector appears
- data types are sent with records queries
- data types are used to enable/disable shot num conversion
- shot nums are considered strings
- `active_area` will appear as `Data Type` instead of `Active Area` as well as have a different icon in the table

All this is to match Gemini functionality whilst still being backwards compatible with EPAC "mode". So this should be tested both against the normal server as well as against the specific Gemini service with the data types configured.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Test with `dataTypes` set to `["GA","GD","GQ","GS"]` and point to Gemini server (ask me for the URL and credentials)
- [x] Test with no `dataTypes` provided against the normal server and verify functionality hasn't changed

## Agile board tracking
Closes DSEGOG-528 Closes DSEGOG-530

### DIFF
--- a/cypress/e2e/channels.cy.ts
+++ b/cypress/e2e/channels.cy.ts
@@ -13,6 +13,35 @@ describe('Data Channels Component', () => {
     cy.get('[role="dialog"]').should('not.exist');
   });
 
+  it('displays active_area as Data Type when in data types mode', () => {
+    let settings = {};
+    cy.request('operationsgateway-settings.json').then((response) => {
+      settings = response.body;
+    });
+
+    cy.intercept('operationsgateway-settings.json', (req) => {
+      req.reply({
+        statusCode: 200,
+        body: {
+          ...settings,
+          dataTypes: ['GS', 'GA', 'GQ', 'GD'],
+        },
+      });
+    }).as('getSettings');
+
+    cy.reload();
+
+    cy.contains('Data Channels').click();
+
+    cy.contains('system').click();
+
+    cy.findByRole('checkbox', { name: 'Data Type' }).check();
+
+    cy.contains('Add Channels').click();
+
+    cy.findByRole('columnheader', { name: 'Data Type' }).should('be.visible');
+  });
+
   it('lets a user add new channels', () => {
     cy.contains('Data Channels').click();
 
--- a/cypress/e2e/search.cy.ts
+++ b/cypress/e2e/search.cy.ts
@@ -949,7 +949,7 @@ describe('Search', () => {
       cy.findByLabelText('date-time search box').should(
         'have.css',
         'border-color',
-        'rgb(211, 47, 47)' // shade of red
+        'rgb(214, 65, 65)' // shade of red
       );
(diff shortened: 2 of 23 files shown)


joshdimanteto (commented):
Works well overall, just a minor styling suggestion.

When switching between data types (many to one, one to many, or one to none), could we disable the transition animation? The change feels a bit noisy during state changes.

https://github.com/user-attachments/assets/7f8da628-beac-454d-a3d4-3987aac1a1c4

I think simply greying out the box is enough to indicate that the input is disabled, without needing the animation.

louise-davies:
@joshdimanteto maybe it seems unnecessary because you don't have any shot numbers visible from the shot number conversion - but the box greying out also clears any existing shot numbers and it un-greying out will display the shot num range for the relevant time stamps.

https://github.com/user-attachments/assets/10d0c1eb-45d1-424f-9394-d36b0d758a3e

Maybe we should disable the animation when it doesn't actually "change" aka we shouldn't highlight from "select" to "select" but I think it clearing out and populating like shown above are still useful?

joshdimanteto:
> @joshdimanteto maybe it seems unnecessary because you don't have any shot numbers visible from the shot number conversion - but the box greying out also clears any existing shot numbers and it un-greying out will display the shot num range for the relevant time stamps.
> Screen.Recording.2026-04-29.105438.mp4
> 
> Maybe we should disable the animation when it doesn't actually "change" aka we shouldn't highlight from "select" to "select" but I think it clearing out and populating like shown above are still useful?

Yeah, I agree with that.

louise-davies:
@joshdimanteto having a look, the issue is with the mock endpoints. If you try this against the actual API - the API returns an error when it can't convert and the animations aren't triggered (as the code correctly deduces that a conversion hasn't happened). The issue is the MSW endpoint doesn't error - it just returns an empty object which our logic then thinks is a valid conversion because it's not `undefined`.

This can be fixed in the code to actually check for the `min`/`max`/`from`/`to` fields to be defined, but I think this is unnecessary and actually the MSW endpoint should error to match what the API returns. Do you agree? If so I'll fix the MSW endpoint

joshdimanteto:
> @joshdimanteto having a look, the issue is with the mock endpoints. If you try this against the actual API - the API returns an error when it can't convert and the animations aren't triggered (as the code correctly deduces that a conversion hasn't happened). The issue is the MSW endpoint doesn't error - it just returns an empty object which our logic then thinks is a valid conversion because it's not `undefined`.
> 
> This can be fixed in the code to actually check for the `min`/`max`/`from`/`to` fields to be defined, but I think this is unnecessary and actually the MSW endpoint should error to match what the API returns. Do you agree? If so I'll fix the MSW endpoint

Yeah, fixing the MSW endpoints to match the API makes more sense

joshdimanteto (approved):
LGTM

PR #854, opened 2026-05-18
DSEGOG-484 Add string type

louise-davies:
## Description
Adds the `string` channel type. Note that `scalar` was previously allowed to be either a number or a string - this was erroneous as `scalar` was always intended to be just numbers so it is now just `number`. This has some knock on effects to `active_area` and `active_experiment`, which are now `string` type, and expands the `dataTypes` configuration to also change the channel type of `shotnum` from `number` to `string` as well. This affects things such as not being able to plot these channels anymore, etc.

Also fixed `Float Image` -> `Float Images` in the export dialogue to match the phrasing of the other options.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Test against gemini API with a channel like `GEM_SHOT_TYPE_STRING` which is a `string` channel

## Agile board tracking
Part of DSEGOG-484

### DIFF
--- a/cypress/e2e/export.cy.ts
+++ b/cypress/e2e/export.cy.ts
@@ -30,8 +30,11 @@ describe('Export', () => {
   it('can select export content', () => {
     cy.findByRole('button', { name: 'Export' }).click();
     cy.findByRole('checkbox', { name: 'Scalars' }).should('be.checked');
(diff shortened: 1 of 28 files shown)


joshdimanteto (approved):
LGTM

PR #896, opened 2026-07-15
DSEGOG-471 Scale images to fit popup window

louise-davies:
## Description
Fixes DSEGOG-471 using option 2, aka "Fit image to popup size when in non-crosshairs mode, and switch to full-size when in crosshairs mode - but limit the max size of the image so it fits in the window and add scrollbars on the image itself to move around".

This is achieved via a lot of CSS manipulation, which I tried to extract out into reuseable functions. Mostly it's a pain because you can't just use CSS - `canvas` requires it's HTML `height` and `width` attributes to be set as well to ensure the scaling is correct.

To not completely mess up the e2e tests, I mostly just adjusted the window size of the popup, so "normal" tests aka ones not testing large iamge display have large popups so the sizing code doesn't kick in, and then there's a seperate test with a "small" sized popup

Ideally test on both Chrome and Firefox as they do behave differently sometimes.

## Testing instructions

- [x] Review code
- [x] Check Actions build
- [x] Test on large images (there should be a column in demo session with a large image column loaded, equally the 12 bit session is also a large enough image to trigger) in both Chrome and Firefox. Try and break it, aka resize the windows, move across monitors, maximize etc.

## Agile board tracking
Closes DSEGOG-471


joshdimanteto (dismissed):
LGTM, just a couple of failing tests

louise-davies:
@joshdimanteto I noticed a few issues with the resize observer not updating properly unless you moved your mouse back into the main OG window - I resolved this by moving the resize observer code to run in the popup, but as this is a significant change could you re-test it yourself as well? Seems to work fine for me, and the code is now simpler because of it, but always good to have two pairs of eyes

joshdimanteto (approved):
LGTM

